PR #20, opened 2020-04-28
ISPN-11733 Report UX changes to caches table

karesti:
Several changes

- [x] Data toolbar instead of Toolbar
- [x] Chips in filters should be displayed. 
- [x] Cache type labels relooked

A warning and prettier code formatter run as well

https://issues.redhat.com/browse/ISPN-11733


pruivo:
integrated! thanks @karesti

PR #28, opened 2020-05-03
ISPN-11490 Fix tabs count task counter

karesti:
Fixes
* Tasks and Counters tab count number display
* The label style for counters and tasks type


tristantarrant:
Merged

PR #29, opened 2020-05-04
ISPN-11705 Global alert system / ISPN-11754 fix cache creation

karesti:
Fixes cache creation when no config is provided (does not create a local cache anymore)

https://issues.redhat.com/browse/ISPN-11754

Alert error 
* Uses useContext
* Custom hook
* For now, one alert at a time
* Create and Delete refactored

https://issues.redhat.com/browse/ISPN-11705


pruivo:
integrated! thanks @karesti !

PR #32, opened 2020-05-06
ISPN-11762 Local activity table

karesti:
https://issues.redhat.com/browse/ISPN-11762


ryanemerson:
@karesti This is really cool :dark_sunglasses: 

One thing that I think is missing though, is that when you press "Clear all entries" the activity table is not updated. I think it would be good if you had something like:

Key: *, Action: Clear, Local Date-Time:

karesti:
@ryanemerson I will rebase and rework this!

ryanemerson:
Great work @karesti , thanks

Issue #16, opened 2020-04-27
Edit entry


karesti:
https://issues.redhat.com/browse/ISPN-11498

https://infinispan.org/docs/dev/titles/rest/rest.html#rest_v2_replace_entries

PR #35, opened 2020-05-07
ISPN-11498 Edit entry

karesti:
https://issues.redhat.com/browse/ISPN-11498

#16

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "webpack-merge": "^4.1.4"
   },
   "dependencies": {
+    "@patternfly/react-catalog-view-extension": "^1.4.58",
     "@patternfly/react-charts": "^5.3.19",
     "@patternfly/react-core": "^3.153.13",
     "@patternfly/react-icons": "^3.15.16",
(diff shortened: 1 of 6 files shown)


pruivo:
It looks good. minor suggestions:

* is it possible to pre-set the Time to live and Max Idle in the edit if previously set?
* is it possible to show the entry after it is created? 
* Time to live and Max Idle is always forever (even if set)
* Looks like the browser caches the fetch requests. If I set max idle and do a search, it never touches the key and it ends up expired
* Created and Last used are never set

karesti:
> * is it possible to pre-set the Time to live and Max Idle in the edit if previously set?

I do agree in all of these. This is related to the addEntry issue that I did not close yet because of the headers issues. There are backend issues yet. I need to check all this with the current master build instead of the latest image.
Maybe would be better to close add/edit and so, and deal with a specific PR for the headers (both in the REST server for eventual corrections and the console)

WDYT ?

pruivo:
ok!
integrated! thanks @karesti

PR #36, opened 2020-05-09
Ignore cache

karesti:
- [x] Refactoring Delete Cache Modal
- [x] Add to ignore cache
- [x] Disable the cache link for Ignore caches
- [x] Put back the ignore cache to the game
- [x] Add a filter for ignore caches => The filter is not under 'Status' filter because the Select grouped with checkbox is bugged. I will contact Patternfly team again. For now I put the 'ignored' filter under 'features' in the select drop down

To make this work, we need the PR of the server in, and then use the master
Infinispan REST API PR: https://github.com/infinispan/infinispan/pull/8301

https://issues.redhat.com/browse/ISPN-11667


karesti:
@pruivo @ryanemerson latest master has the ignore cache endpoint fixed

pruivo:
integrated! thanks @karesti !

PR #40, opened 2020-05-14
ISPN-11821 Land correctly into detail cache

karesti:
https://issues.redhat.com/browse/ISPN-11821

Covers
* Arriving directly to the cache detail from the browse
* Dealing with server errors (not found cache name, or the server down for example)


karesti:
@ryanemerson this is the correction, and using react better

PR #57, opened 2020-06-29
ISPN-12061 Upgrade Patternfly Version

karesti:


pruivo:
integrated. thanks @karesti !

PR #62, opened 2020-07-22
ISPN-12137 Fix protobuf display

karesti:
https://issues.redhat.com/browse/ISPN-12137

### DIFF
--- a/src/services/protobufService.ts
+++ b/src/services/protobufService.ts
@@ -124,7 +124,16 @@ class ProtobufService {
         throw response;
       })
       .then(schema => right(schema))
-      .catch(err => left(<ActionResponse>{ message: err, success: false }));
+      .catch(err => {
+        if (err instanceof TypeError) {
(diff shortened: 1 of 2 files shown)


src/services/protobufService.ts
```diff
@@ -153,7 +162,8 @@ class ProtobufService {
               name: schema.name,
               error: protoError
             };
-          })
+          }).filter(schema => schema.name != 'org/infinispan/protostream/message-wrapping.proto' &&
+            schema.name != 'org/infinispan/query/remote/client/query.proto')
```
ryanemerson, line 166: I'm holding off merging now, as I think this filtering is not needed. The issue is with the Quarkus client. See the #Schemas topic on zulip for more context.

wburns: It will probably be needed until we can get the change added to a released version of Quarkus, which will be a while I am guessing :(

ryanemerson:
Thanks @karesti

PR #67, opened 2020-07-26
ISPN-12168 Upgrade favicons

karesti:
https://issues.redhat.com/browse/ISPN-12168


danberindei:
@karesti it would be useful to have a README or something to explain how those icon names/sizes were chosen. Without it, it's not clear if the icons that were not updated (e.g. `android-chrome-*.png`) should have been updated or removed.

BTW, looks like [we could use SVG icons for Edge/Firefox/Chrome/Chrome for Android](https://caniuse.com/#feat=link-icon-svg). 
Maybe it's worth using only SVG for everything not specific to Apple and Microsoft?

danberindei:
Merged, thanks Katia!

PR #69, opened 2020-08-03
ISPN-12092 Welcome page friendly

karesti:
* Welcome page visible without login
* If Keycloak, when clicking the button, we land into keycloak login
* If Http, we go into the login endpoint
* If http but no user, pop up explaining you need to create a user
* If there is no auth or the server down, go to the console 

https://issues.redhat.com/browse/ISPN-12092


danberindei:
When the server is down I get "TypeError: NetworkError when attempting to fetch resource.", I think we make it even friendlier ;)

src/app/Common/ConsoleBackgroud.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import {BackgroundImage} from '@patternfly/react-core';
+import back1 from '!!url-loader!@app/assets/images/infinispanbg_576.png';
```
danberindei, line 3: Couldn't you import the images as `lg`, `sm` etc., to avoid the extra backX names?

src/app/Support/Support.tsx
```diff
@@ -1,40 +1,74 @@
...
-  EmptyStateIcon,
-  EmptyStateBody,
-  EmptyStateSecondaryActions
-} from '@patternfly/react-core';
+import {Button, Modal, Stack, StackItem, Text, TextContent, TextVariants} from '@patternfly/react-core';
+import {global_FontSize_4xl, global_spacer_sm, global_spacer_md, global_warning_color_100} from "@patternfly/react-tokens";
+import {ExclamationTriangleIcon} from "@patternfly/react-icons";
+
+/**
+ * Support pop up when the user has run the server and there is no user/password to log in
```
danberindei, line 7: @karesti there's always a user/password combination now, the server generates one and logs it in the console:

```
# IDENTITIES_PATH not specified                                                #
# USER and/or PASS env variables not specified.                                #
# Auto generating user and password.                                           #
```

If I cancel the HTTP authentication dialog twice, I don't get this page, I get an empty console instead.
The only message in the browser's console is "Unable to retrieve server version About.tsx:40:16"

src/app/Common/ConsoleBackgroud.tsx
karesti: yes

src/app/Support/Support.tsx
karesti: yep, this is happening when you run a container. running the server without the container is different, you need to create a user/pass and the pop up appears.
let me try what you mention here

karesti: give a try dan please

danberindei:
There's a conflict now @karesti

Have you managed to fix the `TypeError` on Firefox?

karesti:
@danberindei rebased

danberindei:
Error message looks good on Firefox, too.

Merged, thanks Katia!

PR #70, opened 2020-08-03
ISPN-12138 be able to get with other key content types

karesti:
https://issues.redhat.com/browse/ISPN-12138


src/services/utils.ts
```diff
@@ -3,18 +3,17 @@
...
 export enum KeyContentType {
-  OctetStream = 'application/octet-stream',
-  OctetStreamHex = 'application/octet-stream; encoding=hex',
   StringContentType = 'java.lang.String',
   DoubleContentType = 'java.lang.Double',
   IntegerContentType = 'java.lang.Integer',
   LongContentType = 'java.lang.Long',
   BooleanContentType = 'java.lang.Boolean',
-  BytesType = 'Bytes'
+  BytesType = 'Bytes',
```
gustavonalle, line 14: It'd be better to remove the 'Bytes' since it is covered by the other two below

src/services/utils.ts
```diff
@@ -3,18 +3,17 @@
...
+  OctetStream = 'application/octet-stream',
```
gustavonalle, line 15: A less 'technical' way of present this is ```byte[] as Base64```

src/services/utils.ts
```diff
@@ -3,18 +3,17 @@
...
+  OctetStreamHex = 'application/octet-stream; encoding=hex'
```
gustavonalle, line 16: Similar as above, ```byte[] as Hex```

src/services/utils.ts
karesti: ok

src/services/utils.ts
karesti: Which is the  list I should add ? from the console the octet-stream makes sense ?

karesti: and If I add a marshalled 'Person' protostream type, how should I deal it? is it possible ?

gustavonalle: It is not possible to add specific protobuf types to the key content type header. 

> Which is the list I should add 

The same list you have know, but removing ```Bytes``` and adding  ```byte[] as Base64``` and ```byte[] as Hex```

gustavonalle: > Which is the list I should add

You are sending the headers correctly, but the change I suggest is to display it differently (byte[] as Hex)  and (byte[] as Base64)

karesti:
@gustavonalle I changed the display names. btw, if I want to add an Integer instead of a text in the value, how can I do it?

src/services/utils.ts
karesti: great, done and pushed

src/services/utils.ts
karesti: done

gustavonalle:
I can't seen to make it work. 
Steps:
- Create a cache named 'test' with protobuf_dist template
- Add entry
- Key 1
- Value 1

 Error:

Invalid top level object! Found token: NUMBER

gustavonalle:
There is an issue in the server that is causing the edit issues: https://issues.redhat.com/browse/ISPN-12315

gustavonalle:
It needs ISPN-12315, I haven't got to it yet, plan to do it for 12.0.0.Final latest

karesti:
at least the other parts can be integrated if it's working. WDYT ?

gustavonalle:
> at least the other parts can be integrated if it's working. WDYT ?

yes

karesti:
@gustavonalle please grab a look and tell me if it feels like mergeable now (even if some bugs on the server yet)

gustavonalle (approved):

gustavonalle:
Merged, thanks @karesti !

PR #72, opened 2020-08-05
ISPN-12200 Fix the landing into cache configuration templates page

karesti:
To test go directly to http://localhost:9000/console/container/default/configurations/

The app should load correctly

https://issues.redhat.com/browse/ISPN-12200


danberindei:
Merged, thanks Katia!

PR #84, opened 2020-09-28
ISPN-11324 Fix caches manipulation with special characters in names

karesti:
https://issues.redhat.com/browse/ISPN-11324


tristantarrant:
I've tested it locally and it works very well.

PR #85, opened 2020-09-28
[11.x] ISPN-11324 Fix caches manipulation with special characters in names

karesti:
https://issues.redhat.com/browse/JDG-3494


gustavonalle:
Merged, thanks!

PR #101, opened 2020-12-02
[11.x] ISPN-12506 Fix protobuf schema lack of permission

karesti:
https://issues.redhat.com/browse/ISPN-12506

### DIFF
--- a/src/services/protobufService.ts
+++ b/src/services/protobufService.ts
@@ -19,24 +19,25 @@ class ProtobufService {
     schema: string,
     create: boolean
   ): Promise<Either<ActionResponse, ActionResponse>> {
+    // @ts-ignore
     return utils
       .restCallWithBody(
         this.endpoint + '/' + name,
         create ? 'POST' : 'PUT',
         schema
       )
-      .then(response => {
+      .then((response) => {
         if (response.ok) {
           return response.json();
         }
         throw response;
       })
-      .then(schema => {
+      .then((schema) => {
         let protoError: ProtoError | undefined = undefined;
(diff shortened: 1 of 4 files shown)


pruivo:
@karesti I'm unable to test it. I got a blank page even with authentication disabled. better ask someone else.

src/services/protobufService.ts
```diff
@@ -115,58 +81,89 @@ class ProtobufService {
...
+      actionResponse = <ActionResponse>{
+        message: err.message,
+        success: false,
+      };
+    } else if (err instanceof Response) {
+      if (err.status.valueOf() == 403) {
+        // Not Found
+        actionResponse = <ActionResponse>{
+          message:
+            'Unauthorized access. The user lacks ___schema_manager permission.',
```
pruivo, line 153:
```suggestion
            'Unauthorized access.',
```
I don't think it makes sense to leak the permission required. The permission's name is going to change and, using other sites as an example, they just say "unauthorized".

pruivo:
nitpicking... if it isn't too much trouble, maybe "hide" the "Create protobuf schema" if not authorized (or the full tab)?

karesti:
> nitpicking... if it isn't too much trouble, maybe "hide" the "Create protobuf schema" if not authorized (or the full tab)?

Yes. this will be part of a higher auth related feature. We need to do the bare minim here, like we do with other stuff just showing the error. but not dealing with having or not permissions. a well done feature is coming for that

pruivo:
integrated! thanks @karesti !

PR #120, opened 2021-02-01
ISPN-11596 role access console

karesti:
https://issues.redhat.com/browse/ISPN-11596

Needs master infinispan


karesti:
XSite and Query are not working

PR #135, opened 2021-03-16
ISPN-12840 console microcopy edits

oraNod:
https://issues.redhat.com/browse/ISPN-12840

### DIFF
--- a/src/app/assets/languages/en.json
+++ b/src/app/assets/languages/en.json
@@ -24,8 +24,8 @@
     "login-label": "{{brandname}} login.",
     "login": "Log in",
     "cancel": "Cancel",
-    "dialog-label": "Log to Infinispan dialog",
-    "form-username": "User name",
+    "dialog-label": "{{brandname}} log in dialog.",
+    "form-username": "Username",
     "help-username": "Enter your username.",
     "form-password": "Password",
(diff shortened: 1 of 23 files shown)


karesti (approved):

src/app/assets/languages/en.json
```diff
@@ -166,7 +166,7 @@
       "add-entry-form-value-tooltip": "The value can contain simple values ...",
```
oraNod, line 166: @karesti Hi, I've noticed the tooltips for keys and values for the create entry dialog are incomplete. 

"The key can contain simple values ..." and "The value can contain simple values ..." 

How should we describe this? The values for keys and values depends on the encoding, right?

karesti (approved):

src/app/assets/languages/en.json
karesti: yes, we can add json or protobuf entities, depending on the encoding of the cache

oraNod: The key can contain JSON or Protobuf values, depending on cache encoding.

oraNod:
@karesti Hey I updated those two strings. Hopefully the suggestions look OK. Cheers.

PR #141, opened 2021-04-22
ISPN-12458 Index and Query stats REST API

karesti:
* Move the Search code to a Search dedicated file
* Remove deprecated query and index stats calls
* Display the new index and query stats

https://issues.redhat.com/browse/ISPN-12458

### DIFF
--- a/src/services/searchService.ts
+++ b/src/services/searchService.ts
@@ -1,0 +1,286 @@
+import { RestUtils } from './utils';
+import { Either, left, right } from './either';
+import displayUtils from '@services/displayUtils';
+
+/**
+ * Search Service calls Infinispan endpoints related to Search
+ * @author Katia Aresti
+ */
+export class SearchService {
+  endpoint: string;
+  utils: RestUtils;
+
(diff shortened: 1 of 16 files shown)


src/services/searchService.ts
```diff
@@ -0,0 +1,286 @@
...
+    maxResults: number,
+    offset: number
+  ): Promise<Either<ActionResponse, SearchResut>> {
+    return this.utils
+      .restCall(
+        this.endpoint +
+          encodeURIComponent(cacheName) +
+          '?action=search' +
+          '&query=' +
+          query +
```
danberindei, line 38: Shouldn't the query be encoded like the cache name?

danberindei (commented):
I had a quick look, but I'd rather someone who knows more about query merges it

karesti:
I'm going to merge it myself @danberindei because nobody will check any further at this point. if you want to merge it I would love it as well

danberindei:
Good call Katia, I couldn't add anything meaningful.

PR #147, opened 2021-06-02
ISPN-13083 Drop cache templates for local ones

karesti:
https://issues.redhat.com/browse/ISPN-13083

// needs backport

### DIFF
--- a/src/services/dataContainerService.ts
+++ b/src/services/dataContainerService.ts
@@ -119,6 +119,39 @@ export class ContainerService {
   }
 
   /**
+   * Get cache console local configuration templates for in the cache manager name
+   *
+   * @param name, the name of the cache manager
+   */
+  public  getCacheConsoleLocalConfigurationTemplates(): Map<string, string> {
+    let protobuf = <CacheConfig>{
+      name: 'Quick start cache',
+      config: '{\n' +
+        '  "distributed-cache": {\n' +
+        '    "mode": "SYNC",\n' +
+        '    "remote-timeout": 17500,\n' +
+        '    "state-transfer": {\n' +
+        '      "timeout": 60000\n' +
+        '    },\n' +
+        '    "encoding": {\n' +
+        '      "media-type": "application/x-protostream"\n' +
+        '    },\n' +
+        '    "locking": {\n' +
+        '      "concurrency-level": 1000,\n' +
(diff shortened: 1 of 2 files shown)


src/services/dataContainerService.ts
```diff
@@ -117,6 +117,86 @@ export class ContainerService {
       );
   }
 
+  /**
```
karesti, line 120: @tristantarrant which templates should be propose from the console for this version (12) and remove the ones from the server for create cache

karesti: I put that in the code like that but I will probably move this temporal stuff to another file. I want to validate configurations we want to propose for this, and don't any further change in the code.

oraNod: hey, as mentioned in Zulip it seems like the best reason for providing a cache template in the console is to get users off the ground. rather than flood with choices, make it simple and provide one option that works for the getting started use case.

what about keeping only "Distributed Protobuf Encoded Cache" and renaming it to something like "Quick start cache"?

karesti: Good idea @oraNod it can be the replacement for the protobuf example template with no doubt. let's see what @tristantarrant says. I just don't want to confuse users anymore with useless templates. the day we will remove those and deal templates better, we might want to provide a list. but this is work for 13 (I haven't started the fork between 12/13 from the console yet)

oraNod:
@karesti I see these cache templates in the Create cache dialog now: 

image

where do those cache names come from? I think we could possibly change them to make them easier to distinguish/more meaningful.

oraNod:
@karesti deleting my other comment about the cache configuration template names. I was looking in the wrong place. I do think we could make them a little more descriptive so I'd like to improve that before merging.

I guess it's kind of tricky but what I'd like to do is map the template names to suggested use cases.

for example, "Distributed Protobuf Encoded Cache" this is like two keywords from the cache configuration but I'm not sure that helps users choose between the other two. for example, if I'm storing some Java object I might decide to use "Distributed Java Encoded Cache" when in fact Protobuf might be better suited.

also, I'm not sure about "Distributed Java Encoded Cache" because it uses `application/x-java-object` as the encoding. in the docs we state "For best performance, Infinispan recommends storing unmarshalled POJOs in embedded caches only." 

so it's like that one template goes against our own advice.

PR #151, opened 2021-06-16
JDG-4667 Create role can remove caches

karesti:
https://issues.redhat.com/browse/JDG-4667

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2775,6 +2775,16 @@
       "dev": true,
       "optional": true
     },
+    "bindings": {
+      "version": "1.5.0",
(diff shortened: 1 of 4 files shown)


oraNod (approved):
LGTM @karesti I've tested with a user named deployer which is assigned the CREATE role. I can create and delete caches via the console. Is there anything else needed for this or shall we merge?

karesti:
@oraNod ignore caches is back to work. was broken. deployer should not have any possibility to hide or display caches (ignore) but admin should be able to do so

oraNod:
@karesti yep. that looks good to me. deployer cannot show/hide caches but can create/remove. admin can show/hide caches.

PR #177, opened 2022-01-17
ISPN-13510 Cache Metrics size fix

dipguptaredhat:
https://issues.redhat.com/browse/ISPN-13510

### DIFF
--- a/src/app/Caches/CacheMetrics.tsx
+++ b/src/app/Caches/CacheMetrics.tsx
@@ -24,15 +24,31 @@ import {CubesIcon} from '@patternfly/react-icons';
 import {QueryMetrics} from '@app/Caches/Query/QueryMetrics';
 import {ConsoleServices} from "@services/ConsoleServices";
 import {CustomCardTitle} from "@app/Common/CustomCardTitle";
+import {MoreInfoTooltip} from '@app/Common/MoreInfoTooltip';
+import { useTranslation } from 'react-i18next';
 
 const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
   const [stats, setStats] = useState<CacheStats | undefined>(undefined);
   const [statsError, setStatsError] = useState<string | undefined>(undefined);
   const [displayQueryStats, setDisplayQueryStats] = useState<boolean>(false);
+  const [size, setSize] = useState<number|undefined>(0)
+  const [memory, setMemory] = useState<string|undefined>(undefined)
+  const { t } = useTranslation();
 
   useEffect(() => {
     ConsoleServices.caches().retrieveFullDetail(props.cacheName).then((detail) => {
       if (detail.isRight()) {
+
+        // Load the memory storage and size eviction
+        const loadMemory = JSON.parse(detail.value.configuration.config)["distributed-cache"]
+        if(loadMemory.memory){
+          if(loadMemory.memory.storage==="HEAP" && loadMemory.memory["max-size"])
+            setMemory(loadMemory.memory.storage)
+          else if(loadMemory.memory.storage==="OFF_HEAP")
+            setMemory(loadMemory.memory.storage)
+        }
+
+        setSize(detail.value.size)
         setStats(detail.value.stats);
         let loadQueryStats =
           detail.value.stats != undefined && detail.value.stats.enabled && detail.value.features.indexed;
@@ -175,7 +191,7 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
   };
 
   const buildEntriesCard = () => {
-    if (!stats) {
+  if (!stats) {
       return '';
     }
 
@@ -185,17 +201,23 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
         <CardBody>
           <TextContent>
             <TextList component={TextListVariants.dl}>
-              <TextListItem component={TextListItemVariants.dt}>
-                {displayUtils.formatNumber(stats.current_number_of_entries)}
+              <TextListItem aria-label="view-cache-metrics-size" component={TextListItemVariants.dt}>
+                {displayUtils.formatNumber(size)}
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-                Current number
+              <MoreInfoTooltip
+                label={t('caches.cache-metrics.current-number-entries')}
+                toolTip={t('caches.cache-metrics.current-number-entries-tooltip')}
+              />
               </TextListItem>
-              <TextListItem component={TextListItemVariants.dt}>
-                {displayUtils.formatNumber(stats.total_number_of_entries)}
+              <TextListItem aria-label="view-cache-metrics-nodes" component={TextListItemVariants.dt}>
+                {displayUtils.formatNumber(stats.required_minimum_number_of_nodes)}
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-                Total number
+              <MoreInfoTooltip
+              label={t('caches.cache-metrics.min-nodes')}
+              toolTip={t('caches.cache-metrics.min-nodes-tooltip')}
+              />
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>-</TextListItem>
             </TextList>
@@ -216,19 +238,23 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
         <CardBody>
           <TextContent>
             <TextList component={TextListVariants.dl}>
-              <TextListItem component={TextListItemVariants.dt}>
-                {displayUtils.formatNumber(
-                  stats.current_number_of_entries_in_memory
-                )}
+              <TextListItem aria-label="view-cache-metrics-off-heap" component={TextListItemVariants.dt}>
+                { memory==="OFF_HEAP" ? displayUtils.formatNumber(stats.off_heap_memory_used) : "-" }
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-                Current number
+                <MoreInfoTooltip
+                  label={t('caches.cache-metrics.cache-size-off-heap')}
+                  toolTip={memory==="OFF_HEAP" ? '': t('caches.cache-metrics.cache-size-off-heap-tooltip')}
+                />
               </TextListItem>
-              <TextListItem component={TextListItemVariants.dt}>
-                {displayUtils.formatNumber(stats.data_memory_used)}
+              <TextListItem aria-label="view-cache-metrics-heap" component={TextListItemVariants.dt}>
+                { memory==="HEAP" ? displayUtils.formatNumber(stats.data_memory_used) : "-" }
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-                Total in use
+                <MoreInfoTooltip
+                  label={t('caches.cache-metrics.cache-size-heap')}
+                  toolTip={memory==="HEAP" ? '': t('caches.cache-metrics.cache-size-heap-tooltip')}
+                />
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>-</TextListItem>
             </TextList>
@@ -239,7 +265,7 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
   };
 
   if (!stats?.enabled) {
-    const message = 'Statistics disabled.';
+    const message = 'Statistics are disabled.';
 
     return (
       <EmptyState variant={EmptyStateVariant.small}>
--- a/src/app/assets/languages/en.json
+++ b/src/app/assets/languages/en.json
@@ -275,6 +275,16 @@
       "modal-available-description": "will become available for read and write operations.",
       "modal-available-button-done": "OK",
       "modal-available-button-cancel": "Cancel"
+    },
+    "cache-metrics":{
+      "current-number-entries":"Current number of entries",
+      "current-number-entries-tooltip":"Displays the number of entries that are stored in the cache. Each entry is counted only once so the number does not include replicas.",
+      "min-nodes":"Minimum number of nodes",
+      "min-nodes-tooltip":"Displays the number of nodes that must remain in the cluster to avoid data loss.",
+      "cache-size-off-heap":"Size of cache in off-heap memory",
+      "cache-size-off-heap-tooltip":"Displays the amount of off-heap memory that data in the cache uses. Total memory usage is higher because it includes internal data structures.",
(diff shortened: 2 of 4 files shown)


src/app/Caches/CacheMetrics.tsx
```diff
@@ -186,16 +187,16 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
...
               <TextListItem component={TextListItemVariants.dd}>
-                Current number
+                Size
               </TextListItem>
               <TextListItem component={TextListItemVariants.dt}>
-                {displayUtils.formatNumber(stats.total_number_of_entries)}
+                {displayUtils.formatNumber(stats.required_minimum_number_of_nodes)}
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-                Total number
```
karesti, line 198: Please use t hook and externalise the labels into the properties file

karesti (changes requested):
If we can add a cypress IT test to check the size and stats will be great

src/app/Caches/CacheMetrics.tsx
```diff
@@ -186,16 +187,16 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
...
+                Required Minimum No of nodes
```
karesti, line 199: @oraNod Which is the best label for this value?

oraNod: I think "Minimum number of nodes" sounds fine.

src/app/Caches/CacheMetrics.tsx
```diff
@@ -186,16 +187,16 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
           <TextContent>
             <TextList component={TextListVariants.dl}>
               <TextListItem component={TextListItemVariants.dt}>
-                {displayUtils.formatNumber(stats.current_number_of_entries)}
+                {displayUtils.formatNumber(size)}
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-                Current number
+                Size
```
oraNod, line 193: I think we should be more specific here. Size would seem like you mean the total size in some measurement GB/MB. "Current number of entries" is better than "Size" if that's actually what this label shows.

src/app/Caches/CacheMetrics.tsx
```diff
@@ -218,17 +219,17 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
             <TextList component={TextListVariants.dl}>
               <TextListItem component={TextListItemVariants.dt}>
                 {displayUtils.formatNumber(
-                  stats.current_number_of_entries_in_memory
+                  stats.off_heap_memory_used
                 )}
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-                Current number
+                Heap Memory
```
oraNod, line 226: "Amount of heap memory in use" maybe?

src/app/Caches/CacheMetrics.tsx
```diff
@@ -218,17 +219,17 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
...
               </TextListItem>
               <TextListItem component={TextListItemVariants.dt}>
                 {displayUtils.formatNumber(stats.data_memory_used)}
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-                Total in use
+                Data Memory
```
oraNod, line 232: "Amount of memory in use" How is this different to "Amount of heap"? Does this show heap + off-heap?

dipguptaredhat: > "Amount of memory in use" How is this different to "Amount of heap"? Does this show heap + off-heap?

@karesti ?

karesti: @danberindei could you please add some light on these metrics meaning please? :)

danberindei: `data_memory_used` is the amount of heap used to store cache entries, IF the cache is using heap storage, eviction is enabled, and size-based (`<memory storage="HEAP" max-size="some pozitive number"/>`).

`off_heap_memory_used` is the amount of off-heap/native memory used, if the cache is using off-heap storage, regardless of whether eviction is enabled or not (`<memory storage="OFF_HEAP"/>`).

So the labels should probably be "Heap memory" for `data_memory_used` and "Off-heap memory" for `off_heap_memory_used`.

If the console knows whether the cache is using off-heap storage or not, then it should only show the relevant memory stat. If not, then it should display a "N/A" when the stat is missing/-1, and the label (or maybe a tooltip?) could say something like
* `data_memory_used`: Only available when cache is configured to use heap storage and size-based eviction is enabled".
* `off_heap_memory_used`: Only available when cache is configured to use off-heap storage.

Speaking of tooltips, the `data_memory_used` stat only includes memory used to store cache entries, not memory used by internal data structures (e.g. to keep track of transactions, or to keep track of modified keys during state transfer). We might want to have a tooltip even when the stat is present, saying something like "This number only tracks cache entries, total memory usage is higher because of internal data structures that are not tracked".

src/app/Caches/CacheMetrics.tsx
```diff
@@ -216,19 +220,19 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
...
-              <TextListItem component={TextListItemVariants.dt}>
+              <TextListItem aria-label="view-cache-metrics-heap" component={TextListItemVariants.dt}>
                 {displayUtils.formatNumber(
-                  stats.current_number_of_entries_in_memory
+                  stats.off_heap_memory_used
                 )}
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-                Current number
+              {t('caches.cache-metrics.heap-memory')}
```
oraNod, line 229:
```suggestion
              {t('caches.cache-metrics.off-heap-memory')}
```

src/app/assets/languages/en.json
```diff
@@ -268,6 +268,12 @@
       "modal-enable-button" : "Enable",
       "modal-cancel-button" : "Cancel",
       "modal-description" : " Disabling rebalancing prevents the cluster from repairing data in this cache belonging to nodes leaving the cluster. It should be re-enabled as soon as possible to prevent possible data-loss in this cache."
+    },
+    "cache-metrics":{
+      "current-number-entries":"Current number of entries",
+      "min-nodes":"Minimum number of nodes",
+      "heap-memory":"Amount of heap memory in use",
```
oraNod, line 275:
```suggestion
      "off-heap-memory":"Amount of off-heap memory used",
```

src/app/assets/languages/en.json
```diff
@@ -268,6 +268,12 @@
...
+      "data-memory":"Data Memory"
```
oraNod, line 276:
```suggestion
      "data-memory":"Amount of heap memory used"
```

src/app/Caches/CacheMetrics.tsx
oraNod: Thanks @danberindei 

@dipguptaredhat and @karesti What do you think about adding tooltips/hover help here as Dan suggests? I think for both stats we could probably have a single bit of hover help like:

"Displays the total amount of memory that data in the cache uses. The statistic that is available reflects cache configuration, which can be either JVM heap or off-heap, native memory. Total memory usage is higher because it includes internal data structures."

src/app/assets/languages/en.json
```diff
@@ -268,6 +268,12 @@
       "modal-enable-button" : "Enable",
       "modal-cancel-button" : "Cancel",
       "modal-description" : " Disabling rebalancing prevents the cluster from repairing data in this cache belonging to nodes leaving the cluster. It should be re-enabled as soon as possible to prevent possible data-loss in this cache."
+    },
+    "cache-metrics":{
+      "current-number-entries":"Current number of entries",
+      "min-nodes":"Minimum number of nodes",
+      "off-heap-memory":"Amount of off-heap memory used",
+      "data-memory":"Amount of heap memory used"
     }
```
oraNod, line 277: Sorry again @dipguptaredhat But I think we could be more meaningful here yet so a few more nitpicks and small changes. I can't seem to add inline suggestions so I'll just put everything in a single comment. Obvs there are changes here that affect the tsx files too.

How about...

      "off-heap-memory":"Amount of off-heap memory used", -> "cache-size-off-heap":"Size of cache in off-heap memory"
      "data-memory":"Amount of heap memory used" -> "cache-size-heap":"Size of cache in heap memory"

src/app/Caches/CacheMetrics.tsx
karesti: yes we are adding those tooltips

src/app/assets/languages/en.json
karesti: @dipguptaredhat don't forget to resolve the conversation once is done from your side! ;)

src/app/Caches/CacheMetrics.tsx
```diff
@@ -216,19 +221,25 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
...
               </TextListItem>
-              <TextListItem component={TextListItemVariants.dt}>
+              <TextListItem aria-label="view-cache-metrics-heap" component={TextListItemVariants.dt}>
                 {displayUtils.formatNumber(stats.data_memory_used)}
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-                Total in use
+                <MoreInfoTooltip
+                  label={t('caches.cache-metrics.cache-size-heap')}
+                  toolTip={'Choose \'Custom Type\' to specify keys with custom `_type` json value type.'}
```
oraNod, line 241: Change "Choose \'Custom Type\' to specify keys with custom `_type` json value type." to

"Displays the amount of JVM heap memory that data in the cache uses. Total memory usage is higher because it includes internal data structures."

src/app/Caches/CacheMetrics.tsx
```diff
@@ -222,9 +227,7 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
           <TextContent>
             <TextList component={TextListVariants.dl}>
               <TextListItem aria-label="view-cache-metrics-off-heap" component={TextListItemVariants.dt}>
-                {displayUtils.formatNumber(
-                  stats.off_heap_memory_used
-                )}
+                { memory==="OFF_HEAP" ? displayUtils.formatNumber(stats.off_heap_memory_used) : "N/A" }
```
oraNod, line 230: "N/A" does not sound very meaningful to me. I would prefer something like: "This statistic does not apply to this cache."

Do I have the correct understanding here? If this is something we show to users then I think we need to be clear about what we're saying and "N/A" doesn't cut it imo.

oraNod: Thanks for putting it in context to help me understand @dipguptaredhat 

I'd like to get another opinion on the use of "N/A" because I feel like there might be something more intuitive. I'll respond later today so we can close this off.

oraNod: @dvagnero Please also feel free to make a suggestion on the use of "N/A" here. 

This isn't a phrase we currently use elsewhere in the console (see https://github.com/infinispan/infinispan-console/blob/main/src/app/assets/languages/en.json) and - before we adopt it - I'd like to make sure it's the most meaningful option. There might also be a couple of grammar issues. Strictly speaking "n.a." is the correct abbreviation and the translated form is several characters longer than the English or French form. 

The best alternative I can think of is "null" which might be a winner. What are your thoughts?

danberindei: @oraNod I think I suggested "N/A" first, thinking that column is meant to show only numbers, and it's bolded, so putting a longer description there would mess the layout. "N/A" is already a big step up from displaying "-1" or "0" when the number doesn't make sense but the embedded/REST API still provides it as a number.

Maybe "Not Available" would be clearer while still fitting in the column? Then the full explanation could stay in the tooltip and the layout would stay the same.

oraNod: @danberindei Thanks! Yes, I think the full explanation goes in the tooltip. "Not available" is much more meaningful than "N/A" or "n.a." but loses the compactness and messes with the layout. If we translate to Japanese, even if we use "N/A", we'll hit that same issue.

You're spot on regarding "null". That could indicate an error condition. I think "-" is a good option.

oraNod: Let's go with a dash here instead of "N/A" unless there are objections. I've added this in a commit.

oraNod:
@dipguptaredhat Heads up, I've pushed a commit with some suggestions and to add more tooltips.

src/app/assets/languages/en.json
```diff
@@ -271,9 +271,13 @@
     },
     "cache-metrics":{
       "current-number-entries":"Current number of entries",
+      "current-number-entries-tooltip":"Displays the number of entries that are stored in the cache.",
       "min-nodes":"Minimum number of nodes",
+      "min-nodes-tooltip":"Displays the minimum number of nodes for the cache.",
       "cache-size-off-heap":"Size of cache in off-heap memory",
-      "cache-size-heap":"Size of cache in heap memory"
+      "cache-size-heap":"Size of cache in heap memory",
+      "cache-size-heap-tooltip":"Displays the amount of JVM heap memory that data in the cache uses. Total memory usage is higher because it includes internal data structures.",
```
oraNod, line 279: @danberindei How do we measure "amount"? Is this in Gb or Mb or does it depend? I think that's something we need to clarify because until it hits that threshold the console will display 0, even if there are entries in the cache.

src/app/assets/languages/en.json
```diff
@@ -271,9 +271,13 @@
     },
     "cache-metrics":{
       "current-number-entries":"Current number of entries",
+      "current-number-entries-tooltip":"Displays the number of entries that are stored in the cache.",
       "min-nodes":"Minimum number of nodes",
+      "min-nodes-tooltip":"Displays the minimum number of nodes for the cache.",
```
oraNod, line 276: @danberindei What does "minimum number of nodes" reflect? How can we make the tooltip more explanatory?

oraNod:
@dipguptaredhat Here is a commit with additional changes to extract the labels and add tooltips. I've also reviewed the tooltips with Dan. https://github.com/oraNod/infinispan-console/commit/ad79d20d311e4368c0f119137b0bc62388296765

karesti:
@dipguptaredhat needs rebase and merge :)

oraNod (approved):
LGTM

karesti (approved):

PR #185, opened 2022-02-01
ISPN-13651 tooltips for cache metrics

oraNod:
https://issues.redhat.com/browse/ISPN-13651


src/app/Caches/CacheMetrics.tsx
```diff
@@ -186,6 +195,11 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
             />
           </div>
         </CardBody>
+        <CardFooter>
+        <ExpandableSection toggleTextExpanded="Show less" toggleTextCollapsed="Interpreting data access statistics">
+        Data access statistics show info about cache operations to access data.
+        </ExpandableSection>
+        </CardFooter>
```
oraNod, line 202: @karesti Hey, what do you think about adding an expandable footer to the cache metrics donut graph? I'm not sure about how to add tooltips for every data access statistic. Also it seems more compact to add a lot of text to an expandable section rather than a lot of tooltips that might clutter the UI. Just looking for any suggestion on how best to do this. Thanks.

karesti:
@oraNod I think is nice one the expandable

oraNod:
@karesti Thanks. I think it's better than adding all the detail to tooltips and cluttering the donut chart. I've pushed a change and think we're ready for review.

dipguptaredhat (approved):
LGTM

karesti (approved):
@oraNod nice!

PR #186, opened 2022-02-04
ISPN-13684 - Add the wizard use editor option to create caches with a configuration

dipguptaredhat:
* Wizard for cache creation added
* Validation and onNext function pending

### DIFF
--- a/src/app/Caches/Create/CreateCacheWizard.tsx
+++ b/src/app/Caches/Create/CreateCacheWizard.tsx
@@ -1,0 +1,74 @@
+import React, { useEffect, useState } from 'react';
+import { Wizard } from '@patternfly/react-core';
+import GettingStarted from './GettingStarted';
+import { Link } from 'react-router-dom';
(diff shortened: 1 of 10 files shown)


src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -0,0 +1,212 @@
...
+          onBack,
+          onClose,
+        }) => {
+          return (
+            <>
+              <Button
+                variant="primary"
+                type="submit"
+                onClick={() => {
+                  // onNext();
```
dipguptaredhat, line 171: @karesti How can we use onNext() function inside useEffect()?

karesti: you can directly call the method from useEffect. what are you trying to do?

dipguptaredhat: I am not sure how to pass that function in useEffect

PR #191, opened 2022-02-28
ISPN-13684 - Add the wizard use editor option to create caches with a configuration

dipguptaredhat:
* Basic Wizard Structure created

### DIFF
--- a/src/app/Caches/Create/CreateCacheWizard.tsx
+++ b/src/app/Caches/Create/CreateCacheWizard.tsx
@@ -1,0 +1,195 @@
+import React, { useEffect, useState } from 'react';
+import { Wizard } from '@patternfly/react-core';
+import GettingStarted from './GettingStarted';
+import EditCache from './EditConfig';
+import { useHistory } from 'react-router';
+import { useApiAlert } from '@app/utils/useApiAlert';
+import { ConsoleServices } from '@services/ConsoleServices';
+import { CacheConfigUtils } from '@services/cacheConfigUtils';
+import { useTranslation } from 'react-i18next';
+
+const CreateCacheWizard = (props) => {
+    const { addAlert } = useApiAlert();
+    const { t } = useTranslation();
+    const cmName = props.cmName;
+    const [formCacheName, setFormCacheName] = useState('')
+    const [formCreateType, setFormCreateType] = useState<'configure' | 'edit'>('edit');
+    const [formConfig, setFormConfig] = useState('')
+    const [stepIdReached, setStepIdReached] = useState(1);
+    const [isFormValid, setIsFormValid] = useState(false);
+    const [configs, setConfigs] = useState<OptionSelect[]>([]);
+    const [validConfig, setValidConfig] = useState<
+        'success' | 'error' | 'default'
+    >('default');
+    const [errorConfig, setErrorConfig] = useState('');
+    const [selectedConfig, setSelectedConfig] = useState<string>('');
+    const [configExpanded, setConfigExpanded] = useState(false);
+
+    const history = useHistory();
+
+    interface OptionSelect {
+        value: string;
+        disabled?: boolean;
+        isPlaceholder?: boolean;
+    }
+
+    useEffect(() => {
+        ConsoleServices.dataContainer()
+            .getCacheConfigurationTemplates(cmName)
+            .then((eitherTemplates) => {
+                if (eitherTemplates.isRight()) {
+                    let options: OptionSelect[] = [];
+                    eitherTemplates.value.forEach((template) => {
+                        options.push({ value: template.name });
+                    });
+                    setConfigs(options);
+                } else {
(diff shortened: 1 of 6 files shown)


karesti:
@dipguptaredhat the IT test is failing. I will review the PR better just after, but the IT test failing needs to be fixed or removed if it does not apply anymore

src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+        // Validate Name
+        const isValidName: 'success' | 'error' = name.length > 0 ? 'success' : 'error';
+
+        // Validate the config
+        let isValidConfig: 'success' | 'error';
+        let configValidation;
+
+        if (configs.length == 0) {
+            // there are no  templates or the expanded area is opened, we validate the text area content
```
karesti, line 56: create cache here is dealing two things. One of them is creating the cache using a template. This is how it works today. You can leave this behaviour and modify the create cache second step with it (even if it's not in the design right now). To understand how this works, try to run the server with a template defined (check the xml and the it tests) and you will see how it's working today. We can totally keep that behaviour. 

However, what needs to be done is having the parent wizard component keep the state and pass the method that changes the state to the step child component. otherwise, this will never work

karesti (changes requested):

src/app/Caches/Create/EditConfig.tsx
```diff
@@ -0,0 +1,100 @@
...
+        );
+    };
+
+    return (
+        <React.Fragment>
+            <TextContent>
+                <Text component={TextVariants.h3}> {t('caches.create.edit-config.page-title')}</Text>
+            </TextContent>
+            <TextContent style={{ padding: "2rem" }}>
+                <Text component={TextVariants.p}>Lorem ipsum dolor sit amet consectetur adipisicing elit. Laudantium rerum esse fuga ex hic eum accusantium blanditiis perspiciatis omnis. Voluptatem harum tempora animi doloribus tenetur expedita consectetur totam aut molestias?</Text>
```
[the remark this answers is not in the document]
oraNod: @karesti and @dipguptaredhat Just pushed a commit with some suggestions. I want to work with this a bit more closely with @dmvolod and coordinate with the PatternFly UX folks too. So we'll come back to the text strings soon. But I think the commit will be better than lorem ipsum placeholder for now.

src/app/assets/languages/en.json
```diff
@@ -137,16 +137,18 @@
...
+        "cache-name": "1. Enter a unique name for your cache.",
+        "cache-type": "2. Choose how to create your cache.",
+        "cache-type-radio1": "Create a new cache",
+        "cache-type-radio1-helper": "Build a cache with guided steps.",
+        "cache-type-radio2": "Add an existing cache",
+        "cache-type-radio2-helper": "Provide valid cache configuration."
       },
       "edit-config":{
-        "page-title": "Edit code"
+        "nav-title": "Configure ",
```
dipguptaredhat, line 150: @oraNod nav-title is mentioned "Edit Code" on marvel app (https://marvelapp.com/prototype/2b2cgifg/screen/81483820) , Could you please confirm

oraNod: @dipguptaredhat Hey, I don't think we need to strictly follow the wording in the marvel app. I do see that I left out an update in my commit and that string is wrong. I'll fix it now. Thanks for pointing it out and asking.

oraNod: Just pushed another commit to update the wording. I think we'll need to come back to the text strings anyway so let's not regard these as final yet.

src/app/assets/languages/en.json
```diff
@@ -135,7 +135,21 @@
...
-      "back-button-label": "Back"
+      "back-button-label": "Back",
+      "getting-started": {
+        "nav-title": "Get started",
+        "page-title": "Start with the basics",
+        "cache-name": "1. Enter a unique name for your cache.",
+        "cache-type": "2. Choose how to create your cache.",
+        "cache-type-radio1": "Create a new cache",
+        "cache-type-radio1-helper": "Build a cache with guided steps.",
+        "cache-type-radio2": "Add an existing cache",
```
dipguptaredhat, line 146: @karesti @oraNod I am not sure about this text, I think as we are creating a new cache by selecting this radio button. The text should be more related to new cache.

oraNod: @dipguptaredhat Yes, that's a good point. The intention was to say either build a cache from scratch via the wizard or supply an existing cache configuration from a file or some other source. But "existing cache" is not great. I've pushed a change just now.

oraNod (approved):
Great work @dipguptaredhat !!

karesti (approved):

karesti:
@dipguptaredhat fantastic work, I won't nitpick anything at this point. This is a great start and foundation for the wizard feature! I merge and I release the console

PR #194, opened 2022-03-03
Changes for running tests with provided version of server.

andyuk1986:
I have added new profile to pom.xml which is the same as e2e profile except from the part where it compies just compiled console sources to server. The newly created profile is used in case we want to test the existing server with it's own console sources.


karesti (approved):

karesti:
@andyuk1986 thanks!! 👍

PR #196, opened 2022-03-09
ISPN-13685 Improve the wizard to create a cache withtout any feature

dipguptaredhat:
* Progressive Steps in Wizard
* Configuration Basic page created


src/app/Caches/Create/ConfigurationBasic.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+    // Form Number of Owners
+    const formNumberOwners = () => {
+        return (
+            <React.Fragment>
+                <FormGroup
+                    label="Number of owners"
+                    fieldId='field-number-owners'
+                    isRequired={topology === 'distributed'}> {/*Required when topology is distributed */}
+                    <Tooltip content="Tooltip for number of owners"><OutlinedQuestionCircleIcon /></Tooltip>
```
dipguptaredhat, line 142: @karesti How can i place tooltip on right side of the content ?

src/app/Caches/Create/ConfigurationBasic.tsx
```diff
@@ -0,0 +1,231 @@
...
+                    />
+                    <Radio
+                        name="radio"
+                        id="replicated"
+                        onChange={() => setTopology('replicated')}
+                        isChecked={topology === 'replicated'}
+                        label={
+                            <TextContent>
+                                <Text component={TextVariants.h3}>Replicated</Text>
+                            </TextContent>
```
karesti, line 111: @dipguptaredhat labels need to be used. there are some content constant with this values

src/app/Caches/Create/ConfigurationBasic.tsx
```diff
@@ -0,0 +1,231 @@
...
+    // Form Number of Owners
+    const formNumberOwners = () => {
+        return (
+            <React.Fragment>
+                <FormGroup
+                    label="Number of owners"
+                    fieldId='field-number-owners'
+                    isRequired={topology === 'distributed'}> {/*Required when topology is distributed */}
+                    <Tooltip content="Tooltip for number of owners"><OutlinedQuestionCircleIcon /></Tooltip>
+                    <Select
```
karesti, line 143: use numberinput 
https://www.patternfly.org/v4/components/number-input

src/app/Caches/Create/ConfigurationBasic.tsx
```diff
@@ -0,0 +1,231 @@
...
+                        {optionsNumberOwners}
+                    </Select>
+                </FormGroup>
+            </React.Fragment>
+        )
+    }
+
+    // Options for Encoding Cache
+    const optionsEncodingCache = [
+        <SelectOption key={0} value="JSON" isPlaceholder />,
```
karesti, line 161: the encoding options are not ok.
the encoding type is in InfinispanRefData.EncodingType. Same for the type of caches

src/app/Caches/Create/ConfigurationBasic.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+                    <TextContent>
+                        <Text component={TextVariants.p}>Lorem ipsum dolor sit amet consectetur adipisicing elit. Corporis ratione eveniet necessitatibus? Consequuntur excepturi itaque fugiat maiores quas, tempora sunt quod distinctio deleniti quis iusto dolorem optio nam eligendi sequi!</Text>
+                    </TextContent>
+                </FormGroup>
+
+                {formTopology()}
+
+                {/* Display the number of owners of the cache when the topology is distributed. */}
+                {topology === 'distributed' && formNumberOwners()}
```
karesti, line 221: use ref data

src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -25,6 +59,11 @@ const CreateCacheWizard = (props) => {
     const [selectedConfig, setSelectedConfig] = useState<string>('');
     const [configExpanded, setConfigExpanded] = useState(false);
 
+    const [stepIdReached, setStepIdReached] = useState(1);
+
+    // State for the form (Configuration Basic)
+    const [formConfigurationBasic, setFormConfigurationBasic] = useState({ "formTopology": '', "numberOfOwners": '', "encodingCache": '', "isStatistics": false });
```
karesti, line 65: probably this should be exported as a type in the file InfinispanTypes.ts

src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -151,6 +185,70 @@ const CreateCacheWizard = (props) => {
...
+                        ...stateObj,
+                        showEditCodeStep: true,
+                        showConfigurationStep: false,
+                        showReviewStep: false
+                    },
+                    () => callback()
+                );
+            }
+        }
+        else if (activeStep.name === t('caches.create.edit-config.nav-title')) {
```
karesti, line 220: I don't think is a good idea to use the translator here. steps should have an id that is not depending of the name. and then use the translator only to display

src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -177,18 +275,80 @@ const CreateCacheWizard = (props) => {
...
+        name: 'Configuration',
+        steps: [
+            {
+                id: 3,
+                name: 'Basic',
+                component: (
+                    <ConfigurationBasic formConfigurationBasic={formConfigurationBasic} handleFormConfigurationBasic={handleFormConfigurationBasic} />
+                )
+            },
+            { id: 4, name: 'Cache Feature', component: <p>Cache Feature</p> },
```
karesti, line 288: try to already export these into the translator file. DOn't use <p>, use text content component

src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -177,18 +275,80 @@ const CreateCacheWizard = (props) => {
...
+                {({ activeStep, goToStepByName, goToStepById, onNext, onBack, onClose }) => {
+                    return (
+                        <>
+                            <Button
+                                variant="primary"
+                                type="submit"
+                                onClick={(event) => getNextStep(event, activeStep, onNext)}
+                                isDisabled={!isFormValid}
+                            >
+                                {activeStep.name === t('caches.create.edit-config.nav-title') || activeStep.name === 'Review' ? "Create Cache" : "Next"}
```
karesti, line 319: here too, don't use the translator this way

src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -177,18 +275,80 @@ const CreateCacheWizard = (props) => {
...
+                            </Button>
+                            <Button
+                                variant="secondary"
+                                onClick={(event) => getPreviousStep(event, activeStep, onBack)}
+                                className={
+                                    activeStep.name === t('caches.create.getting-started.nav-title') ? "pf-u-hidden" : ""
```
karesti, line 325: classe names are not used hardcoded this way. you need to use a patternfly token

src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -177,18 +275,80 @@ const CreateCacheWizard = (props) => {
...
+                                }
+                            >
+                                Back
+                            </Button>
+                            <Button variant="link" onClick={onClose}>
+                                Cancel
+                            </Button>
```
karesti, line 332: export these already please

karesti (changes requested):
some changes already requested

karesti:
@dipguptaredhat the radio button should be already selected to the first option in the form. Radio buttons are always preselected with the most common choice, so we don't force always the user to chose something IMO

src/app/Caches/Create/ConfigurationBasic.tsx
dipguptaredhat: Labels are already in use, could you please elaborate.

src/app/Caches/Create/ConfigurationBasic.tsx
```diff
@@ -116,51 +129,43 @@ const ConfigurationBasic = ({ formConfigurationBasic, handleFormConfigurationBas
...
+            </React.Fragment >
         )
     }
 
     // Options for Encoding Cache
     const optionsEncodingCache = [
         <SelectOption key={0} value="JSON" isPlaceholder />,
-        <SelectOption key={1} value="YAML" />,
-        <SelectOption key={2} value="XML" />,
+        <SelectOption key={1} value="Java" />,
```
dipguptaredhat, line 161: @karesti value here is mentioned as a string, instead of EncodingType ref from InfinispanRefData.ts, Is this the right way? not sure!

src/app/Common/MoreInfoTooltip.tsx
```diff
@@ -1,22 +1,26 @@
...
   global_FontSize_sm,
   global_spacer_xs,
+  global_FontWeight_bold
 } from '@patternfly/react-tokens';
 
 /**
  * This component is used to add tooltips in the labels of the forms
  */
-const MoreInfoTooltip = (props: { label: string; toolTip: string }) => {
+const MoreInfoTooltip = (props: { label: string; toolTip: string, textComponent?: TextVariants }) => {
```
dipguptaredhat, line 23: @karesti @oraNod Added two new props (optional). Please verify if they don't disrupt other things.

src/app/assets/languages/en.json
```diff
@@ -150,6 +152,31 @@
...
         "page-title": "Provide cache configuration",
         "template-help-invalid": "Select a template or enter a configuration"
+      },
+      "configurations":{
+        "nav-title": "Configuration",
+        "basic":{
+          "nav-title": "Basic",
+          "page-title": "Basic configuration",
+          "topology-title": "Topology of the clustered cache",
+          "topology-tooltip": "Select the topology of the clustered cache.",
```
dipguptaredhat, line 162: @oraNod Please verify the wordings

src/app/Caches/Create/ConfigurationBasic.tsx
karesti: Replicated is not using the constant, the value is written in the code

src/app/Caches/Create/ConfigurationBasic.tsx
karesti: you can display the EncodigType as string

karesti:
@dipguptaredhat added a commit

src/app/Caches/Create/Review.tsx
```diff
@@ -0,0 +1,192 @@
...
+        // Update the form when the state changes
+        props.reviewOptionsModifier({
+            createCacheCode: createCacheCode,
+            downloadType: downloadType
+        });
+    }, [createCacheCode, downloadType]);
+
+    /*TODO: Create similar config for Replicated cache */
+    // Distributed Cache config with variables
+    const distributedCache = {
```
dipguptaredhat, line 51: @karesti If it's the right way to create cache config?

karesti: no. I will add a commit

src/services/infinispanRefData.ts
```diff
@@ -83,3 +83,18 @@ export enum InfinispanFlags {
   SKIP_XSITE_BACKUP = 'SKIP_XSITE_BACKUP',
   ZERO_LOCK_ACQUISITION_TIMEOUT = 'ZERO_LOCK_ACQUISITION_TIMEOUT',
 }
+
+export enum EvictionStrategy {
+  MaximumCount = 'MaximumCount',
```
karesti, line 88: values are NONE, MANUAL, REMOVE, EXCEPTION

karesti (changes requested):

src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -121,123 +80,27 @@ const CreateCacheWizard = (props) => {
...
-                    cacheEditor.configs = options;
-                } else {
-                    addAlert(eitherTemplates.value);
-                }
-            })
-    }, []);
-
     const title = 'Create Cache';
 
-    const createCache = () => {
```
dipguptaredhat, line 76: @karesti this function was working for creating cache with templates also. Aren't we going to use templates for cache creation in future also?

src/app/Caches/Create/AdvancedOptions.tsx
```diff
@@ -0,0 +1,300 @@
...
+import { useTranslation } from 'react-i18next';
+import { MoreInfoTooltip } from '@app/Common/MoreInfoTooltip';
+
+const AdvancedOptions = (props: {
+    advancedOptions: AdvancedConfiguration,
+    advancedOptionsModifier: (AdvancedOptions) => void,
+}) => {
+    const { t } = useTranslation();
+
+    const [storage, setStorage] = useState<"heap" | "off_heap">(props.advancedOptions.storage || "heap");
```
karesti, line 28: extract heap and off_heap into constants ?

src/app/Caches/Create/AdvancedOptions.tsx
```diff
@@ -0,0 +1,300 @@
...
+        setConcurrencyLevel(value);
+    }
+
+    const handleLockAcquisitionTimeout = (value) => {
+        setLockAcquisitionTimeout(value);
+    }
+
+    const onChangeMaxSize = (value, inputValue, setLocalInputValue) => {
+        let newValue;
+        if (inputValue === undefined) {
```
karesti, line 89: extract to a utility method and add unit test

src/app/Caches/Create/AdvancedOptions.tsx
```diff
@@ -0,0 +1,300 @@
...
+                newValue = Math.floor(inputValue);
+            }
+        }
+        setMaxSize(newValue);
+        setInputValueMaxSize(newValue)
+    };
+
+    const onChangeMaxCount = (value, inputValue, setLocalInputValue) => {
+        let newValue;
+        if (inputValue === undefined) {
```
karesti, line 108: DRY with the method above

you can change this to !inputValue instead of === undefined

src/app/Caches/Create/AdvancedOptions.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+    return (
+        <React.Fragment>
+            <Form>
+                {formMemory()}
+                <Divider />
+                {formLocking()}
+
+            </Form>
+        </React.Fragment>
```
karesti, line 296: you don't need React.Fragment if you have a begin and close tag. With <\Form\> is enough

src/app/Caches/Create/ConfigurationBasic.tsx
```diff
@@ -0,0 +1,213 @@
...
+                    />
+                </FormGroup>
+            </React.Fragment>
+        );
+    };
+
+    // Form Number of Owners
+    const formNumberOwners = () => {
+        return (
+            <React.Fragment>
```
karesti, line 123: remove React.Fragment

src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -157,38 +176,134 @@ const CreateCacheWizard = (props) => {
...
+        createCacheCall
+            .then((actionResponse) => {
+                if (actionResponse.success) {
+                    history.push('/');
+                }
+                return actionResponse;
+            })
+            .then((actionResponse) => addAlert(actionResponse));
+
+        // ConsoleServices.caches().createCacheWithConfiguration(
```
karesti, line 282: remove commented code

src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -157,38 +176,134 @@ const CreateCacheWizard = (props) => {
...
-        <Wizard
-            navAriaLabel={`${title} steps`}
-            mainAriaLabel={`${title} content`}
-            onNext={onNext}
-            onBack={onBack}
-            onClose={closeWizard}
-            onSave={onSave}
-            steps={steps}
-        />
+        <>
```
karesti, line 296: remove this tag

karesti (changes requested):
some more comments

src/app/Caches/Create/AdvancedOptions.tsx
dipguptaredhat: Instead of extracting to utility function, i have created a new helper function in the same file. Would be okay?

karesti: for  unit testing if it's in another file

karesti:
@dipguptaredhat I've reviewed the behaviour. The behaviour of the option working now (giving a xml yaml or json config and having an example json and the templates) is broken now. 

This is not ok

import { CacheEditor } from "@app/Caches/Create/CreateCacheWizard";

const CacheEditor = (props: {
    cacheEditor: CacheEditor,
    cacheEditorModifier: (CacheEditor) => void,
}) 

I don't know what happened in this part of the code, but the steps should be separarte. And having something called CacheEditor as a property with the same name of the component CacheEditor must be changed.

karesti:
Another thing I think we must change is the expiration part. We put that in the advanced options but it's a cache : "bounded". The expiration option should not be there

src/services/cacheConfigUtils.ts
```diff
@@ -146,4 +146,51 @@ export class CacheConfigUtils {
...
+          },
+          locking: {
+            isolation: data.advanced.isolationLevel,
+          },
+          memory: {
+            'max-count': data.advanced.maxCount,
+            'max-size': data.advanced.maxSize,
+            'when-full': data.advanced.evictionStrategy,
+          },
+          indexing: {
```
dipguptaredhat, line 187: @karesti Got an error when creating cache with Replicated Topology, regarding indexing storage 

Screenshot from 2022-03-24 12-18-40

dvagnero:
Hi @dipguptaredhat, while working on the docs part I noticed some undesired behavior. 
1. The Cache name text field allows me to enter the same cache name twice. Steps to reproduce: 1. Have a cache named hello. 2. Click Create a cache and in the Cache name field enter hello again. The Cache name text field shows a green checkmark. If I continue to the Review page and hit the Create button I get an error message that the cache hello already exists. My point is that we shouldn't allow customers to proceed with invalid cache name. 
2. It's also possible to proceed without providing any Cache name. Steps to reproduce: 1. Enter a cache name. 2. Delete the cache name. The Cache name text field keeps the green checkmark even though there is no text. 3. Click the next button and continue to the Review page where you get the error message (screenshot attached). Again I think we should notify customers right away that the cache name is invalid.
3. It's recommended to use switch for enabling/disabling some functionality. Could you use switch for enabling Statistics and Striping? Here is some info regarding that https://www.patternfly.org/v4/components/switch/design-guidelines 

Screenshots for #2
Screenshot 2022-03-23 at 14-39-11 Caches

Screenshot 2022-03-23 at 14-36-40 Caches

Please let me know if you need any more info.

src/services/cacheConfigUtils.ts
karesti: the indexing part should not be present

karesti:
@dipguptaredhat would you mind squasing all the commits so I do a new fresh review ?

src/services/cacheConfigUtils.ts
tristantarrant: > Another thing I think we must change is the expiration part. We put that in the advanced options but it's a cache feature : "bounded". The expiration option should not be there. When we will add the features we will put that back. Sorry about that

Careful: Bounded is Eviction, not Expiration.

src/services/cacheConfigUtils.ts
```diff
@@ -146,4 +147,132 @@ export class CacheConfigUtils {
 
     return contentTypes;
   }
+
+  public static createCacheConfigFromData(data: CacheConfiguration): string {
+    let cache;
+
+    if (data.basic.topology === 'Distributed') {
+      cache = {
+        'distributed-cache': {
```
dipguptaredhat, line 156: Storage (heap or off_heap) is not added in the configuration yet. Not sure about that, because I saw `"storage" : "local-heap"` also in the documentation.

karesti (changes requested):
The encoding "EMPTY" should not be displayed.
```json
"encoding": {
      "media-type": "Empty"
    },
    "lock
```

encoding empty means no encoding but we don't want to encourage this from the wizard, so please remove it as an option

karesti:
@dipguptaredhat The review page should display the cache name that is going to be created IMO, since we have already forgotten about it too. 
Would you mind adding Review and the cache name in the title ?

karesti:
Screenshot 2022-03-29 at 10 53 02

src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -1,154 +1,153 @@
...
 
-        if (selectedConfig != '') {
-            // User has chosen a template
-            isValidConfig = 'success';
-        } else {
-            if (configs.length == 0 || configExpanded) {
-                // there are no  templates or the expanded area is opened, we validate the text area content
-                configValidation = CacheConfigUtils.validateConfig(formConfig);
+    const onGoToStep = (id, prevId) => {
+        // console.log(id, prevId);
```
karesti, line 93: remove

src/services/cacheConfigUtils.ts
karesti: @oraNod might help more on this

karesti (changes requested):
a couple of minor comments

karesti:
@dipguptaredhat this is almost ready, good work!! :))))

src/services/fetchCaller.ts
```diff
@@ -97,6 +97,27 @@ export class FetchCaller {
     );
   }
 
+  /**
+   * Get REST API Calls
+   *
+   * @param url
+   */
+  public head(url: string): Promise<Either<any, any>> {
```
dipguptaredhat, line 105: @karesti Head api created but not working, please check.

dipguptaredhat:
> Hi @dipguptaredhat, after yesterdays update I noticed that the Cache name field preserves only the first letter of cache name. Steps to reproduce when creating a cache:
> 
> 1. Enter cache name that has more than one letter :)
> 2. Click Next to continue with configuration
> 3. Click Back to return to the Get started page. The Cache name filed preserves only the first letter. The issue occurs also in the scenario when I proceed with configuration and create cache.

Was a minor bug, it's solve you can check now!

oraNod:
@karesti @dipguptaredhat @dvagnero I think we should come up with a term to describe the wizard itself. I don't think "wizard" is the right choice. It sounds dated probably difficult for translation. 

What do you think about "Cache builder"?

image

image

src/services/cacheConfigUtils.ts
oraNod: @dipguptaredhat `local-heap` is different. That configures how Infinispan stores the index for a cache. It stores the index in the local JVM heap instead of somewhere persistent so it's volatile. 

For cache configuration, we should only be concerned with heap / off-heap. This is where Infinispan stores entries in the cache, either in the JVM heap (managed memory space) or off-heap (native, unmanaged memory).

Note that it's always "off-heap" with a hyphen. If you haven't seen it you can find that in the terminology guide at: https://infinispan.org/docs/stable/titles/contributing/contributing.html#terminology

karesti:
This is almost ready for merge. We need to work the tooltips and the descriptions. we still have some lorem ipsun dolor around @dipguptaredhat @oraNod

oraNod:
@karesti I'm working on the tooltips and wording now. Will have the commit real soon. If it suits better we can do docs separately but right after this is merged. Depends on what you and @dipguptaredhat think.

src/app/Caches/Create/AdvancedOptions.tsx
```diff
@@ -0,0 +1,183 @@
...
+                            </TextContent>
+                        }
+                    />
+                </FormGroup>
+            </React.Fragment>
+        );
+    }
+
+    // Options for Isolation Level
+    const isolationLevelOptions = () => {
```
oraNod, line 108: @dipguptaredhat and @karesti I think the console should show only the supported options.

Need to remove/hide "SERIALIZABLE" and "READ_UNCOMMITTED" from the drop-down. Should we do that in this PR or open a separate issue?

@pruivo FYI can we remove these options from the schema as well?

oraNod: Actually there should only be two options for the isolation level: READ_COMMITTED and REPEATABLE_READ. NONE makes no sense for clustered caches.

pruivo: @oraNod TBH, I want to remove the `isolation-level` attribute :rofl:
See point 2 in https://issues.redhat.com/browse/ISPN-3927
+1 to remove the unsupported options

dvagnero:
@dipguptaredhat I added a commit with docs updates from @oraNod and me

karesti (approved):

PR #197, opened 2022-03-23
Support http RBAC (not only Digest)

karesti:


andyuk1986:
Hi,
I have tested the changes with BASIC and Digest authentication methods and all work as expected. Thanks a lot for your work!

PR #200, opened 2022-03-30
ISPN-13687 Feature Bounded

dipguptaredhat:
https://issues.redhat.com/browse/ISPN-13687


karesti (changes requested):
The cache bounded relates to all the part of the eviction strategy, max entries and so on. 
The select box should not be in capital letters neither.
I merged with main.

Please check the docs and the word document. I think we can do this feature first completely and then move to the next, like the persistence one.

karesti:
@dipguptaredhat rebase from main

src/services/cacheConfigUtils.ts
```diff
@@ -154,53 +154,54 @@ export class CacheConfigUtils {
...
+        (cacheType = distributedCache))
+      : ((cache = { [replicatedCache]: generalCache }),
+        (cacheType = replicatedCache));
+
+    const expiration = () => {
+      cache[cacheType]['expiration'] = {
+        lifespan: data.basic.lifeSpan,
+        'max-idle': data.basic.maxIdle,
+
+        // Todo: Find Proper variable name
```
dipguptaredhat, line 188: @karesti @oraNod For reaper and wake up interval, what would be the key name? can't find on documentation.

oraNod: @dipguptaredhat Could it be here? https://github.com/infinispan/infinispan/blob/main/core/src/main/java/org/infinispan/configuration/cache/ExpirationConfiguration.java#L18

dipguptaredhat:
@oraNod @dvagnero please review the  labels.

karesti:
@oraNod @dvagnero labels review is needed now. thanks!!!

dvagnero:
I'll check it this afternoon. @karesti @oraNod

src/app/Caches/Create/ConfigurationFeature.tsx
```diff
@@ -0,0 +1,209 @@
...
+    };
+
+    // Options for Max Size Unit
+    const unitOptions = () => {
+        return Object.keys(MaxSizeUnit).map((key) => (
+            <SelectOption key={key} value={MaxSizeUnit[key]} />
+        ));
+    }
+
+    const formBoundedCache = () => {
```
karesti, line 136: extract this to another component in another file

src/app/Caches/Create/ConfigurationFeature.tsx
```diff
@@ -0,0 +1,209 @@
...
+            setMaxSize(undefined);
+            setMaxSizeNumber(undefined);
+        }
+        if (maxCount === '') setMaxCount(undefined);
+    }, [maxSizeNumber, maxSizeUnit, maxCount])
+
+    useEffect(() => {
+        props.cacheFeatureModifier({
+            cacheFeatureSelected: cacheFeatureSelected,
+            maxSize: maxSize,
```
karesti, line 66: max size can be set to negatif number and in this case cache creation fails.
the max count gets disabled too

karesti (changes requested):
there is some weird behaviour when using bounded caches.
I believe we should be able to chose in the wizard if we want o set up in milliseconds or seconds hours or days the expiration time, puting seconds by default (same as memory where we chose MG KB etc). And then convert the number in the cache configuration to the chosen milliseconds.

karesti:
Screenshot 2022-04-11 at 16 58 48

karesti:
@dipguptaredhat I don't like that in bounded caches when I chose the feature to have error messages, and then one of those gets disabled. I think we need to provide a way to say that either we chose max size either we chose max number, but not both.

karesti:
Screenshot 2022-04-13 at 15 25 18

karesti:
@dipguptaredhat I think we need a radio button to chose between maxSize or maxCount and then display one or the other.
Will be easier to understand 
https://www.patternfly.org/v4/components/radio

src/app/Caches/Create/ConfigurationFeature.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+    const [isOpenCacheFeature, setIsOpenCacheFeature] = useState(false);
+    const [isOpenEvictionStrategy, setIsOpenEvictionStrategy] = useState(false);
+    const [isOpenMaxSizeUnit, setIsOpenMaxSizeUnit] = useState(false);
+
+    // Helper states for the maxSize input
+    const [maxSizeNumber, setMaxSizeNumber] = useState<string>();
+    const [maxSizeUnit, setMaxSizeUnit] = useState<MaxSizeUnit>(MaxSizeUnit.MB);
+
+    useEffect(() => {
```
dipguptaredhat, line 46: @karesti UseEffect here is used multiple times, not sure if it needs refracting.

karesti: should be ok.

karesti:
@dipguptaredhat I think we can move forward and rework the rewording and that's all

karesti:
@dipguptaredhat rebase the commits and I merge!! thanks!

oraNod:
@karesti Hi, I spoke to Dipanshu earlier and am working on some string updates. I'll have them ready today. Can we wait for the merge? I'll try to be quick.

src/app/Caches/Create/ConfigurationFeature.tsx
```diff
@@ -0,0 +1,275 @@
...
+                </Select>
+            </FormGroup>
+        )
+    }
+
+    // Options for Eviction Strategy
+    const evictionStrategyOptions = () => {
+        return Object.keys(EvictionStrategy).map((key) => (
+            <SelectOption key={key} value={EvictionStrategy[key]} />
+        ));
```
oraNod, line 158: @dipguptaredhat Hey, one comment about the eviction strategy drop-down. I think it should show only REMOVE and EXCEPTION. It doesn't make sense to set NONE as the eviction strategy because that means eviction won't happen so is kind of a misconfiguration. And MANUAL is for embedded caches only. From the server there is no way for users to manually invoke `evict()` on a remote cache.

oraNod (approved):
I've pushed some changes to the strings and tooltips. Also one comment about the eviction strategy drop-down. LGTM otherwise.

Great work @dipguptaredhat

karesti (approved):

PR #207, opened 2022-04-21
ISPN 13688 feature indexed

dipguptaredhat:
Feature Indexed cache added.


src/services/infinispanRefData.ts
```diff
@@ -135,3 +135,8 @@ export enum EvictionType {
   size = 'size',
   count = 'count',
 }
+
+export enum IndexedStorage {
```
dipguptaredhat, line 139: Storage type "Persistent" & "volatile" not working at last step

src/services/cacheConfigUtils.ts
```diff
@@ -203,9 +207,54 @@ export class CacheConfigUtils {
       };
     };
 
+    // Helper for Index entities
+    const indexedEntitiesHelper = () => {
+      let indexedEntities = [];
+      data.feature.indexedCache.indexedEntities.forEach((entity) => {
+        indexedEntities.push('indexed-entity' + ':' + entity);
+      });
+      return indexedEntities;
```
dipguptaredhat, line 216: Screenshot from 2022-04-22 00-30-36

indexed-entities is getting as an array, instead we need as a object.

karesti:
@dipguptaredhat I merged bounded and rebased main and pushed to be easier on the review and focused on indexing only.

karesti:
@dipguptaredhat I added a commit to show you a way to use the editor changes to create the cache config. One thing to be done is when you go back cleanup the reviewed config I think. I let you test and fix whatever looks weird.

Concerning indexing, the search field is not possible today. We don't have and endpoint (at least for now) thats lists the entities you can add to indexing. Needs to be done by hand.

The indexing options are too complicated too. I think we need to let the user configure a couple of thing only and if it needs to change something then should go and change it. -1 in every field is very confusing too IMO. We need to decide what is really important to be configured on the wizard and what can be let out in all these options. @fax4ever might be able to help on that.

karesti:
@dipguptaredhat for indexing entities we need an input to type the indexing entity
https://www.patternfly.org/v4/components/text-input

and then display them as removable label or chip 
https://www.patternfly.org/v4/components/label
https://www.patternfly.org/v4/components/chip

src/app/Caches/Create/CreateCacheWizard.tsx
```diff
@@ -52,9 +52,38 @@ const BoundedCacheInitialState: BoundedCache = {
...
+    queueCount: -1,
+    queueSize: -1,
+    ramBufferSize: -1,
+    threadPoolSize: -1,
+}
+
+const IndexMergeInitialState: IndexMerge = {
+    calibrateByDeletes: false,
+    factor: -1,
+    maxEntries: -1,
```
karesti, line 68: Needs to be reviewed. I don't know if we want to let all this to be configured this way

src/app/Caches/Create/Features/IndexedCache.tsx
```diff
@@ -0,0 +1,396 @@
...
+        return (
+            <FormGroup fieldId='indexed-entities'>
+                <MoreInfoTooltip label="Indexed Entities" toolTip="Indexed Entities Tooltip" textComponent={TextVariants.h2} />
+                <TextInputGroup>
+                    <TextInputGroupMain
+                        value={entityInput}
+                        onChange={(val) => setEntityInput(val)}
+                        onKeyDown={handleTextInputKeyDown}
+                        placeholder="search"
+                        aria-label="Search input"
```
karesti, line 129: As I explained, can't be a search. You need to provide a textfield where the user will type de value (for example com.redhat.PurchaseOrderItem), click on a button "add" and display them as label or chip so it can be removed or changed. I think label might be better as they are editable now too (beta feature) https://www.patternfly.org/v4/components/label

karesti (changes requested):
see my comments

src/app/Caches/Create/CreateCacheWizard.tsx
dipguptaredhat: @oraNod as [suggested](https://infinispan.zulipchat.com/#narrow/stream/118645-infinispan/topic/Cache.20Wizard/near/279398697) , added the options for tuning indexing, let us know if all the options are necessary or we can remove some.

src/app/Caches/Create/Features/IndexedCache.tsx
dipguptaredhat: To add item, user can enter after entering in text field, should we also need "add" button?

src/services/infinispanRefData.ts
dipguptaredhat: @oraNod in the cache config, this storage type is not working, what should be the perfect wording for this?

src/app/Caches/Create/CreateCacheWizard.tsx
oraNod: Just to be clear with my proposal above the wizard would have this navigation:

- Basic options 
- Infinispan capabilities 
- Cache tuning options

oraNod: Right now I think the indexing section is pretty spot on, we just need to move all the tuning parameters out of there.

src/services/infinispanRefData.ts
oraNod: @dipguptaredhat https://docs.jboss.org/infinispan/14.0/apidocs/org/infinispan/configuration/cache/IndexStorage.html 

I believe it's

```
export enum IndexedStorage {
  persistent = 'filesystem',
  volatile = 'local_heap',
}
```

karesti:
@dipguptaredhat I think the labels on the top are not great. IMO the indexed entities should be listed below.
Concerning indexing options, as I said, there are too many. This needs to be cleaned up. The config can be changed in the editor after, but I don't think we should display so many fine grained tuning options on indexing. The goal is not being able to configure absolutely everything, but those options that are usually changed or configured and help users to 

This interface options are really too complex.

Screenshot 2022-05-01 at 14 13 03
do so.

karesti:
@oraNod @fax4ever would you mind checking briefly these options and let those that are usually tuned? We can also add a link in the console to send to the configuration part on the docs that will send the user to other properties if needed. I just believe there are too many options and makes this interface is too complex, too quickly and unnecessarily in this wizard.

src/app/assets/languages/en.json
```diff
@@ -231,7 +229,35 @@
           "striping": "Lock striping",
           "striping-tooltip": "Uses a shared pool of locks for all entries in the cache. Striping lowers the memory footprint for locks but can reduce concurrency. If you disable striping, a lock is created for each entry in the cache.",
           "index-tuning": "Index tuning",
-          "index-tuning-tooltip": "Control internal components that {{brandname}} uses for interacting with an index."
+          "index-tuning-tooltip": "Control internal components that {{brandname}} uses for interacting with an index.",
```
dipguptaredhat, line 232: @oraNod I added some labels from [Documentation](https://infinispan.org/docs/stable/titles/query/query.html#index_reader) , please review

dipguptaredhat:
@karesti squashed all commits in one.

src/app/Caches/Create/Features/IndexedCache.tsx
```diff
@@ -0,0 +1,166 @@
...
+                            </Label>
+                        ))}
+                    </StackItem>
+                </Stack>
+            </FormGroup >
+        )
+    }
+
+    return (
+        <Form onSubmit={(e) => {
```
dipguptaredhat, line 119: @karesti Form submit issue is still there under index entities field, please review.

karesti:
@dipguptaredhat I added a commit that fixes 
The indexing in advanced options is still buggy. 
The merge indexing is included despite the fact I did not configure anything.
The form for the advanced options is messily arranged. Check how I added some changes please.

The 2nd commit fixes the undefined issue. For undefined we can use ? with typescript. On replicated we can have undefined in some cases so typescript raises the issue when +1 or -1 to undefined. This can be avoided with the code I added and we are protected against runtime issues.

dipguptaredhat:
@karesti added a commit for defining undefined on Striping which is a boolean value. The problem with this was defining a default value on Switch and undefined won't seem a good fit. So tried making the switch "disabled" and enable when the user clicks on the switch first time.Though this is not the feasible solution, will need to look more on defining default values of such boolean fields.

src/types/InfinispanTypes.ts
```diff
@@ -337,17 +337,18 @@ interface CacheFeatureStep {
...
-  concurrencyLevel: number | undefined;
-  isolationLevel: string | undefined;
-  lockAcquisitionTimeout: number | undefined;
-  striping: boolean;
+  storage?: string;
+  concurrencyLevel?: number;
+  isolationLevel?: string;
+  lockAcquisitionTimeout?: number;
+  striping?: boolean;
   indexReader: number | undefined;
```
karesti, line 345: aren't these not mandatory too ?

karesti (changes requested):

src/types/InfinispanTypes.ts
dipguptaredhat: We have to pass those as an optional (?) parameter, because in the end if they doesn't have a value it won't show in the Cache Config.

dipguptaredhat:
> The form for the advanced options is messily arranged. Check how I added some changes please.

Do we also need Cards in Advance option page?

src/app/Caches/Create/Features/IndexedCache.tsx
```diff
@@ -0,0 +1,150 @@
...
+        setValidEntity('success')
+      } else {
+        setValidEntity('error');
+      }
+    }
+  }
+
+  return (
+    <Form onSubmit={(e) => {
+      e.preventDefault();
```
dipguptaredhat, line 69: @karesti Prevent default is not working, When we press enter on Indexed entities, the page get refreshed.

PR #212, opened 2022-05-10
ISPN-13690 Feature form for secured caches

dipguptaredhat:
Feature secured cache added
https://issues.redhat.com/browse/ISPN-13690

### DIFF
--- a/src/app/Caches/Create/Features/SecuredCache.tsx
+++ b/src/app/Caches/Create/Features/SecuredCache.tsx
@@ -1,0 +1,120 @@
+import React, { useState, useEffect } from 'react';
+import {
+    Card,
+    CardBody,
+    CardHeader,
+    Form,
+    FormGroup,
+    Label,
+    Select,
+    SelectOption,
+    SelectVariant,
+    Text,
+    TextContent,
+    TextVariants,
+} from '@patternfly/react-core';
+import { useTranslation } from 'react-i18next';
+import { ConsoleServices } from '@services/ConsoleServices';
+import { global_spacer_sm } from '@patternfly/react-tokens';
+
+const SecuredCache = (props: {
+    securedOptions: SecuredCache,
+    securedOptionsModifier: (SecuredCache) => void,
+}) => {
+
+    const { t } = useTranslation();
+    const brandname = t('brandname.brandname');
+
+    const [roles, setRoles] = useState<string[]>(props.securedOptions.roles);
+
+    const [loading, setLoading] = useState(true);
+    const [availableRoles, setAvailableRoles] = useState<string[]>([]);
+    const [error, setError] = useState<'success' | 'error' | 'default'>('default');
+
+    const [isOpenRoles, setIsOpenRoles] = useState(false);
+
+    useEffect(() => {
+        if (loading) {
+            ConsoleServices.security()
+                .getSecurityRoles()
+                .then((r) => {
+                    if (r.isRight()) {
+                        setAvailableRoles(r.value);
+                        setError('success')
+                    }
+                    else {
+                        setError('error');
+                    }
+                }).then(() => setLoading(false));
+        }
+    }, [loading]);
+
+    useEffect(() => {
+        props.securedOptionsModifier({
+            roles: roles,
+        });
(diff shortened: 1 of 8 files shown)


src/services/cacheService.ts
```diff
@@ -621,4 +621,15 @@ export class CacheService {
       errorMessage: `Cache ${cacheName} already exists.`,
     });
   }
+
+  /**
+   * Retrieve security roles
+   *
+   */
+  public async getSecurityRoles(): Promise<Either<ActionResponse, string[]>> {
```
karesti, line 629: move to securityService
this method is unrelated to caches

src/app/Caches/Create/Features/SecuredCache.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    const onSelectRoles = (event, selection) => {
+        if (roles.includes(selection))
+            setRoles(roles.filter(role => role !== selection));
+        else
+            setRoles([...roles, selection]);
+    }
+
+    return (
+        <Form>
```
karesti, line 72: check the Card and CardBody changes I reported in indexing to change this accordingly.

src/app/Caches/Create/Features/SecuredCache.tsx
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }, [loading]);
+
+    useEffect(() => {
+        props.securedOptionsModifier({
+            roles: roles,
+        });
+    }, [roles]);
+
+    // Options for Isolation Level
```
karesti, line 56: delete comment

src/app/Caches/Create/Features/SecuredCache.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const SecuredCache = (props: {
+    securedOptions: SecuredCache,
+    securedOptionsModifier: (SecuredCache) => void,
+}) => {
+
+    const { t } = useTranslation();
+    const brandname = t('brandname.brandname');
+
+    //Secured Cache
```
karesti, line 25: delete this comment, not necessary information

karesti (changes requested):
some changes that depend from the changes I added / asked in the indexing PR

dipguptaredhat:
@karesti new commit added, in the last step create cache not working, please check

karesti:
@dipguptaredhat rebase needed from main! ping me DM when ready please!

karesti:
@dipguptaredhat choosing a role should be mandatory for secured caches next. otherwise we are at the end with a configuration that is not correct

```json
 "security": {
      "authorization": {
        "enabled": true,
        "roles": []
      }
    }
```

karesti:
@dipguptaredhat there was a bug in 13.0.10.Final and 14.x parsing the configuration of secured caches. works with 13.0.10.Final but it's not fixed in 14.0.0.Dev02
https://github.com/infinispan/infinispan/commit/82c1e00a2c7c6b8cc9e4950b9efde13eac5712a6

oraNod:
@karesti and @dipguptaredhat Hey, what do you think about changing this to "RBAC" instead of "secured"?

When I think of security I think of SSL/TLS so changing the label to RBAC might be more immediately descriptive to users. They'd know what to expect when they select the option.

I ask this because I can change the labels but they won't necessarily correspond to Infinispan types, etc. So I would propose that if we make this change we do a bit of refactoring to s/secured/rbac. I know that's a pain though. Sorry I didn't mention it earlier but I've been caught up in stuff.

tristantarrant:
I'm not sure the RBAC acronym is well-known enough. I'd think that ACL is much more common.

oraNod:
Maybe we could even go with "authorization" to be consistent with the schema?

karesti:
@oraNod make any doc/label change you need. I vote for authorization too

karesti (changes requested):
Rename header label for Authorization instead of secured

Chose at least 1 role before moving forward.

dipguptaredhat:
@karesti changes done and squashed

karesti (approved):

PR #218, opened 2022-06-06
ISPN-13750 Add download option for configuration

dipguptaredhat:
Download button on review page, for downloading cache config with json, yaml or xml format.


src/services/cacheService.ts
```diff
@@ -621,4 +621,61 @@ export class CacheService {
...
+      header2,
+      ContentTypeHeaderMapper.fromContentType(contentType2)
+    );
+    return customHeaders;
+  }
+
+  /**
+   * Convert cache configuration to XML or JSON or YAML
+   */
+  public async convertConfigFormat(
```
dipguptaredhat, line 660: @karesti I created a API for converting cache format, but not working properly, please check.

karesti (changes requested):
I pushed a commit with the fix

The download button is not very adapted in the place where it's now. Could we try to put it next to the create button instead ?

karesti:
Sorry I changed your commit instead of crating a new one. 

Feel free to report the changes I did and force push with the change I asked please

src/services/fetchCaller.ts
```diff
@@ -170,7 +170,7 @@ export class FetchCaller {
     return response
       .then((response) => {
         if (response.ok) {
-          return 'response.text()';
```
karesti, line 173: the issue was here

src/services/fetchCaller.ts
```diff
@@ -186,6 +186,7 @@ export class FetchCaller {
         return <ActionResponse>{
           message: successMessage,
           success: true,
+          data: message
```
karesti, line 189: I added a way to take the body into the action response

dipguptaredhat:
@karesti Download button is now on the right side of the footer, let me if that works fine.

PR #219, opened 2022-06-07
ISPN-13691 Feature form for persisted caches

dipguptaredhat:
Feature form for Persisted Cache


src/app/utils/persistentStorageTemplate.ts
```diff
@@ -0,0 +1,54 @@
...
+  "dialect": "H2",
+  "shared": "true",
+  "table-name": "books",
+  "schema": {
+    "message-name": "books_value",
+    "package": "library"
+  }
+}`;
+
+export const jdbcStore = ``;
```
dipguptaredhat, line 42: Can't find example for JDBC Storage type, and please if you could verify all other templates

src/app/utils/persistentStorageTemplate.ts
```diff
@@ -0,0 +1,54 @@
+export const fileStore = `"file-store": {
+  "shared": false,
```
ryanemerson, line 2: The default value for "shared" is false, so we can remove this from the config.

src/app/utils/persistentStorageTemplate.ts
```diff
@@ -0,0 +1,54 @@
...
+  "data": {
+    "path": "data"
+  },
+  "index": {
+    "path": "index"
+  },
+  "write-behind": {
```
ryanemerson, line 9: Adding the "write-behind" object sets the store to be async, which most users won't want as the default behaviour. We should remove this.

src/app/utils/persistentStorageTemplate.ts
```diff
@@ -0,0 +1,54 @@
...
+  "remote-server": [
+    {
+      "host": "one",
+      "port": "12111"
+    },
+    {
+      "host": "two"
+    }
+  ],
+  "connection-pool": {
```
ryanemerson, line 26: We can remove this object and let the user use the default connection-pool configuration.

src/app/utils/persistentStorageTemplate.ts
```diff
@@ -0,0 +1,54 @@
...
+    "path": "index"
+  },
+  "write-behind": {
+    "modification-queue-size": "2048"
+  }
+}`;
+
+export const remoteStore = `"remote-store": {
+  "cache": "mycache",
+  "raw-values": "true",
```
ryanemerson, line 16: The default value is "false" and is applicable for most use-cases, so we should remove this.

src/app/utils/persistentStorageTemplate.ts
```diff
@@ -0,0 +1,54 @@
...
+  "remote-server": [
+    {
+      "host": "one",
```
ryanemerson, line 19: I know these values are from the docs, but we should use an ip address for the hosts. I would also just include a single server, as I think this will fit most use-cases. The definition being inside an array implies that more than one object can be defined.

src/app/utils/persistentStorageTemplate.ts
ryanemerson:
```json
"string-keyed-jdbc-store": {
		"dialect": "H2",
		"string-keyed-table": {
			"prefix": "ISPN_STRING_TABLE",
			"create-on-start": true,
			"id-column": {
				"name": "ID_COLUMN",
				"type": "VARCHAR(255)"
			},
			"data-column": {
				"name": "DATA_COLUMN",
				"type": "BINARY"
			},
			"timestamp-column": {
				"name": "TIMESTAMP_COLUMN",
				"type": "BIGINT"
			},
			"segment-column": {
				"name": "SEGMENT_COLUMN",
				"type": "INT"
			}
		},
		"connection-pool": {
			"connection-url": "jdbc:h2:mem:infinispan",
			"driver": "org.h2.Driver",
			"username": "sa",
			"password": "changeme"
		}
	}
```

src/app/utils/persistentStorageTemplate.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const sqlStore = `"table-jdbc-store": {
+  "dialect": "H2",
+  "shared": "true",
+  "table-name": "books",
+  "schema": {
+    "message-name": "books_value",
+    "package": "library"
+  }
+}`;
```
ryanemerson, line 40: I think we need to include the "connection-pool" here to indicate to the user that this must be configured.

```json
	"table-jdbc-store": {
		"dialect": "H2",
		"shared": "true",
		"table-name": "books",
		"connection-pool": {
			"connection-url": "jdbc:h2:mem:infinispan",
			"driver": "org.h2.Driver",
			"username": "sa",
			"password": "changeme"
		}
	}
```

src/app/utils/persistentStorageTemplate.ts
```diff
@@ -0,0 +1,54 @@
...
+export const jdbcStore = ``;
+
+export const rocksDB = `"rocksdb-store": {
+  "path": "rocksdb/data",
+  "expiration": {
+    "path": "rocksdb/expired"
+  }
+}`;
+
+export const clusterLoader = `"cluster-loader" : {
```
ryanemerson, line 51: IMO we shouldn't include the ClusterLoader.

src/services/infinispanRefData.ts
```diff
@@ -140,3 +140,13 @@ export enum IndexedStorage {
   persistent = 'filesystem',
   volatile = 'local_heap',
 }
+
+export enum PersistentCacheStorage {
+  FileStore = 'File Store',
+  RemoteStore = 'Remote Store',
+  SQLStore = 'SQL Store',
+  JDBCStore = 'JDBC Store',
```
ryanemerson, line 148: The name should be `'JDBC String Based Store'' in order to distinguish it from the `SQLStore` which is also utilises JDBC.

ryanemerson (changes requested):
My review has focused on the templates. I haven't looked at the Typescript code.

IMO the templates used in the Wizard should provide the minimal configuration required for the store to function in it's most common use-case. I imagine a lot of users will attempt to utilise the templates without modification, so it's important that we don't include "extra" configuration such as "write-behind" that aren't applicable to the most common use-case.

src/app/utils/persistentStorageTemplate.ts
dipguptaredhat: Should the cache be like this? with a single host

```
"remote-store": {
  "cache": "mycache",
  "remote-server": [
    {
      "host": "one",
      "port": "12111"
    },
  ]
}
```

ryanemerson: You just need to remove the last comma:

```json
"remote-store": {
  "cache": "mycache",
  "remote-server": [
    {
      "host": "one",
      "port": "12111"
    }
  ]
}
```

ryanemerson: > I know these values are from the docs, but we should use an ip address for the hosts.

I should have updated my example to include ^. Also, the default port for Hot Rod is `11222` so we should utilise that in the template:

```json
"remote-store": {
  "cache": "mycache",
  "remote-server": [
    {
      "host": "127.0.0.1",
      "port": "11222"
    }
  ]
}
```

dipguptaredhat:
@ryanemerson Branch rebased and the last suggestion was also done :+1:

ryanemerson (approved):

dipguptaredhat:
@ryanemerson got this error while creating persistent cache with "rocks-db" storage
Screenshot from 2022-06-15 11-24-16

config:
```
{
  "distributed-cache": {
    "mode": "SYNC",
    "owners": 1,
    "statistics": true,
    "encoding": {
      "media-type": "application/x-protostream"
    },
    "locking": {
      "isolation": "REPEATABLE_READ"
    },
    "persistence": {
      "rocks-db": {
        "path": "rocksdb/data",
        "expiration": {
          "path": "rocksdb/expired"
        }
      }
    }
  }
}
```

src/app/Caches/Create/FeaturesSelector.tsx
```diff
@@ -59,20 +63,24 @@ const FeaturesSelector = (props: {
...
-            validForm = backupsFeatureValidation(validForm);
-            validForm = transactionalFeatureValidation(validForm);
+            const boundedValid = boundedFeatureValidation(validForm);
+            const indexingValid = indexingFeatureValidation(validForm);
+            const securedValid = securedFeatureValidation(validForm);
+            const backupsValid = backupsFeatureValidation(validForm);
+            const transactionalValid = transactionalFeatureValidation(validForm);
+            const persistentValid = persistentFeatureValidation(validForm);
+
+            validForm = boundedValid && indexingValid && securedValid && backupsValid && transactionalValid && persistentValid;
```
dipguptaredhat, line 79: @karesti Made some changes in feature validation, now if any of the feature validation is false, the final would be false.

ryanemerson:
> got this error while creating persistent cache with "rocks-db" storage

I think the store parent element should be `"rocksdb-store"`

karesti:
@dipguptaredhat @oraNod I simplified after speaking with Ryan since having more than one store is VERY UNLIKE TO HAPEN. In general 1 cache = 1 store, so configuring multiple stores is confusing and complicated.
I also simplified the templates since the "table-store" and so on are part of the persistent store config and should not be in the end but in the editor as an example

karesti (approved):

PR #225, opened 2022-06-22
ISPN-13962 Refactoring to adapt to PTF4 forms

karesti:
https://issues.redhat.com/browse/ISPN-13962


karesti:
@dipguptaredhat @andyuk1986 @oraNod don't do any change to the wizard until I'm over here please

oraNod (commented):
@karesti I pushed a few minor doc suggestions. One thing that really sticks out to me is the fact the console seems to throw an error as soon as you select a capability.

image

From a user's perspective it's somewhat alarming. It's like you think you did something wrong and it takes a second.

I feel like it would be better if this popped up when selecting the "Next" button if there are errors. Otherwise the user shouldn't see this imo.

karesti:
@oraNod before you could not click on next and you did not know what you missed in any feature. We need to make it visible to the users fixes the errors. We can launch the validation on next button if you wish, but the original form did not enable the button. I did not change anything, I just made  the errors disabling the next button visible. Until now you just need to figure out by yourself without any hint on error displayed.

karesti:
We can add a warning error on the top for now. And decide if we change the next button behaviour. But we can't do both (disable next button and not display any error)

oraNod (approved):

karesti:
@oraNod your commit is gone but not your changes ;)

PR #231, opened 2022-07-18
ISPN-11659 Data distribution for less than 5 nodes

dpanshug:
https://issues.redhat.com/browse/ISPN-11659

Chart can be viewed using options:
- Total number of entries
- Total memory used

Remaining: applying threshold value

### DIFF
--- a/src/app/Caches/CacheMetrics.tsx
+++ b/src/app/Caches/CacheMetrics.tsx
@@ -1,4 +1,4 @@
-import React, {useEffect, useState} from 'react';
+import React, { useEffect, useState } from 'react';
 import {
   Card,
   CardBody,
@@ -22,21 +22,22 @@ import {
   ListItem
 } from '@patternfly/react-core';
 import displayUtils from '@services/displayUtils';
-import {ChartDonut, ChartThemeColor} from '@patternfly/react-charts';
-import {CubesIcon} from '@patternfly/react-icons';
-import {QueryMetrics} from '@app/Caches/Query/QueryMetrics';
-import {ConsoleServices} from "@services/ConsoleServices";
-import {CustomCardTitle} from "@app/Common/CustomCardTitle";
-import {MoreInfoTooltip} from '@app/Common/MoreInfoTooltip';
+import { ChartDonut, ChartThemeColor } from '@patternfly/react-charts';
+import { CubesIcon } from '@patternfly/react-icons';
+import { QueryMetrics } from '@app/Caches/Query/QueryMetrics';
+import { DataDistribution } from '@app/Caches/DataDistribution';
+import { ConsoleServices } from "@services/ConsoleServices";
+import { PopoverHelp } from '@app/Common/PopoverHelp';
 import { ExpandableSection } from '@patternfly/react-core';
 import { useTranslation } from 'react-i18next';
+import { StorageType } from "@services/infinispanRefData";
 
 const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
   const [stats, setStats] = useState<CacheStats | undefined>(undefined);
   const [statsError, setStatsError] = useState<string | undefined>(undefined);
   const [displayQueryStats, setDisplayQueryStats] = useState<boolean>(false);
-  const [size, setSize] = useState<number|undefined>(0)
-  const [memory, setMemory] = useState<string|undefined>(undefined)
+  const [size, setSize] = useState<number | undefined>(0)
+  const [memory, setMemory] = useState<string | undefined>(undefined)
   const { t } = useTranslation();
   const brandname = t('brandname.brandname');
 
@@ -80,28 +81,34 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
                 {displayUtils.formatNumber(stats.average_read_time)}
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-              <MoreInfoTooltip
-                label={t('caches.cache-metrics.average-reads')}
-                toolTip={t('caches.cache-metrics.average-reads-tooltip')}
-              />
+                <PopoverHelp
+                  name='average-reads'
+                  label={t('caches.cache-metrics.average-reads')}
+                  content={t('caches.cache-metrics.average-reads-tooltip')}
+                  text={t('caches.cache-metrics.average-reads')}
+                />
               </TextListItem>
               <TextListItem component={TextListItemVariants.dt}>
                 {displayUtils.formatNumber(stats.average_write_time)}
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-              <MoreInfoTooltip
-                label={t('caches.cache-metrics.average-writes')}
-                toolTip={t('caches.cache-metrics.average-writes-tooltip')}
-              />
+                <PopoverHelp
+                  name='average-writes'
+                  label={t('caches.cache-metrics.average-writes')}
+                  content={t('caches.cache-metrics.average-writes-tooltip')}
+                  text={t('caches.cache-metrics.average-writes')}
+                />
               </TextListItem>
               <TextListItem component={TextListItemVariants.dt}>
                 {displayUtils.formatNumber(stats.average_remove_time)}
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-              <MoreInfoTooltip
-                label={t('caches.cache-metrics.average-deletes')}
-                toolTip={t('caches.cache-metrics.average-deletes-tooltip')}
-              />
+                <PopoverHelp
+                  name='average-deletes'
+                  label={t('caches.cache-metrics.average-deletes')}
+                  content={t('caches.cache-metrics.average-deletes-tooltip')}
+                  text={t('caches.cache-metrics.average-deletes')}
+                />
               </TextListItem>
             </TextList>
           </TextContent>
@@ -115,7 +122,7 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
       return;
     }
     return (
-      <QueryMetrics cacheName={props.cacheName}/>
+      <QueryMetrics cacheName={props.cacheName} />
     );
   };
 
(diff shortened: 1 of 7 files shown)


dipguptaredhat:
@tristantarrant Can you specify from where we can get the `threshold` value?

src/app/Caches/CacheMetrics.tsx
```diff
@@ -79,28 +80,28 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
                 {displayUtils.formatNumber(stats.average_read_time)}
               </TextListItem>
               <TextListItem component={TextListItemVariants.dd}>
-              <MoreInfoTooltip
-                label={t('caches.cache-metrics.average-reads')}
-                toolTip={t('caches.cache-metrics.average-reads-tooltip')}
-              />
+                <MoreInfoTooltip
```
karesti, line 83: Use PopoverHelp

src/app/Caches/CacheMetrics.tsx
```diff
@@ -136,19 +137,19 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
       <Card>
         <CardTitle>{t('caches.cache-metrics.data-access-title')}</CardTitle>
         <CardBody>
-          <div style={{height: '300px', width: '70%'}}>
+          <div style={{ height: '300px', width: '70%' }}>
```
karesti, line 140: extract to scss file as a class

src/app/Caches/CacheMetrics.tsx
```diff
@@ -136,19 +137,19 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
...
               constrainToVisibleArea={true}
               data={[
-                {x: 'Hits', y: stats.hits},
-                {x: 'Misses', y: stats.misses},
-                {x: 'Stores', y: stats.stores},
-                {x: 'Retrievals', y: stats.retrievals},
-                {x: 'Remove Hits', y: stats.remove_hits},
-                {x: 'Removes Misses', y: stats.remove_misses},
-                {x: 'Evictions', y: stats.evictions},
+                { x: 'Hits', y: stats.hits },
```
karesti, line 144: extract to labels

src/app/Caches/CacheMetrics.tsx
```diff
@@ -263,21 +264,21 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
           <TextContent>
             <TextList component={TextListVariants.dl}>
               <TextListItem aria-label="view-cache-metrics-off-heap" component={TextListItemVariants.dt}>
-                { memory==="OFF_HEAP" ? displayUtils.formatNumber(stats.off_heap_memory_used) : "-" }
+                {memory === "OFF_HEAP" ? displayUtils.formatNumber(stats.off_heap_memory_used) : "-"}
```
karesti, line 267: use constant

src/app/Caches/CacheMetrics.tsx
```diff
@@ -288,6 +289,10 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
     );
   };
 
+  const buildDataNodes = () => {
+    return <DataDistribution cacheName={props.cacheName} />;
```
karesti, line 293: no need a method for this

src/app/Caches/CacheMetrics.tsx
```diff
@@ -315,6 +320,7 @@ const CacheMetrics = (props: { cacheName: string; display: boolean }) => {
       <GridItem span={4}>{buildEntriesCard()}</GridItem>
       <GridItem span={4}>{buildMemoryCard()}</GridItem>
       <GridItem span={4}>{buildOperationsPerformanceCard()}</GridItem>
+      <GridItem span={12}>{buildDataNodes()}</GridItem>
```
karesti, line 323: use the component directly

src/app/Caches/DataDistribution.tsx
```diff
@@ -0,0 +1,121 @@
...
+    Spinner,
+} from '@patternfly/react-core';
+import { Chart, ChartBar, ChartGroup, ChartVoronoiContainer } from '@patternfly/react-charts';
+import { TableErrorState } from '@app/Common/TableErrorState';
+import { useTranslation } from 'react-i18next';
+import { CustomCardTitle } from "@app/Common/CustomCardTitle";
+import { useDataDistribution } from '@app/services/dataDistributionHook';
+
+const DataDistribution = (props: {
+    cacheName: string;
```
karesti, line 20: check if we need this parameter or we can use the context provider of cache detail and a hook to retrieve all

src/app/Caches/DataDistribution.tsx
```diff
@@ -0,0 +1,121 @@
...
+        const data = dataDistribution?.map((item) => {
+            const yaxis = statsOption === "Total number of entries" ? item.total_entries : item.memory_entries;
+            return { name: item.node_name, y: yaxis, x: item.node_name }
+        }
+        );
+
+        return (
+            <div style={{ height: '450px', width: '700px', margin: "auto" }}>
+                <Chart
+                    ariaDesc="Average number of pets"
```
karesti, line 57: extract labels

src/app/services/dataDistributionHook.ts
```diff
@@ -0,0 +1,27 @@
...
+  const [dataDistribution, setDataDistribution] = useState<
+    DataDistribution[]
+  >();
+  const [error, setError] = useState('');
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    if (loading) {
+      ConsoleServices.caches()
+        .getDistribution()
```
karesti, line 14: You need the cache name

src/services/cacheService.ts
```diff
@@ -678,4 +678,43 @@ export class CacheService {
       body: config,
     });
   }
+
+  public async getDistribution(): Promise<
+    Either<ActionResponse, DataDistribution[]>
```
karesti, line 683: you need the cache name

karesti (changes requested):
you can use Dev04 server now that has been released

src/app/Caches/DataDistribution.tsx
dipguptaredhat: I guess we need the parameter. It was passed from routes to parent components of this component.

dipguptaredhat:
@jabolina For using endpoint at `/rest/v2/cluster?action=distribution`, do we need to pass any header?
https://issues.redhat.com/browse/ISPN-12193

src/app/Caches/CacheMetrics.tsx
dipguptaredhat: Do we have a scss file already where we can export?

karesti:
@dipguptaredhat the endpoint you are using is for a global distribution of data, not for a cache. 
https://infinispan.org/docs/14.0.x/titles/rest/rest.html#rest_v2_cache_distribution

dpanshug:
@andyuk1986 can you review this, thanks

karesti:
I will review

src/services/cacheService.ts
```diff
@@ -678,4 +678,50 @@ export class CacheService {
...
+    cacheName: string
+  ): Promise<Either<ActionResponse, DataDistribution[]>> {
+    return this.fetchCaller.get(
+      this.endpoint +
+        '/caches/' +
+        encodeURIComponent(cacheName) +
+        '?action=distribution',
+      (text) => text
+    );
+    // const dataDistrib = <DataDistribution[]>[
```
karesti, line 692: remove commented code

src/services/infinispanRefData.ts
```diff
@@ -210,3 +210,9 @@ export enum ConfigDownloadType {
   XML = 'XML',
   YAML = 'YAML',
 }
+
+export enum DataDistributionStatsOption {
+  Entries = 'Total number of entries',
```
karesti, line 215: This should be in the translation so the enum can be used that way

karesti (changes requested):
I will check the behaviour

karesti:
If we don't have the Total number of entries in memory, then we need to remove it

karesti:
Screenshot 2022-08-22 at 12 35 00

karesti:
@dpanshug seems a bit odd the display

I also realized that while the distribution endpoint  and others are not over, nothing is displayed in the tab. There is a message of statistics not enabled. however, if statistics are enabled but the endpoints are still pending of response from the server, the display of each part of the tab of stats should display the spinner

dpanshug:
> @dpanshug seems a bit odd the display
> 
> I also realized that while the distribution endpoint and others are not over, nothing is displayed in the tab. There is a message of statistics not enabled. however, if statistics are enabled but the endpoints are still pending of response from the server, the display of each part of the tab of stats should display the spinner

Yes we should add a spinner when data is loading, but can show Distribution not available when no nodes or entries are present?

karesti:
> > @dpanshug seems a bit odd the display
> > I also realized that while the distribution endpoint and others are not over, nothing is displayed in the tab. There is a message of statistics not enabled. however, if statistics are enabled but the endpoints are still pending of response from the server, the display of each part of the tab of stats should display the spinner
> 
> Yes we should add a spinner when data is loading, but can we display Distribution not available when no nodes or entries are present?

how could no nodes be present? For the entries part, you can display a message with there are not entries or something like that yes.

src/services/infinispanRefData.ts
dpanshug: Using translation there is throwing an error:
`Uncaught Error: Invalid hook call. Hooks can only be called inside of the body of a function component`

and also enum does not allow computed values.

karesti: you can't use translation there, you need to be in the react component. you could use a label name instead

src/services/infinispanRefData.ts
```diff
@@ -210,3 +210,8 @@ export enum ConfigDownloadType {
   XML = 'XML',
   YAML = 'YAML',
 }
+
+export enum DataDistributionStatsOption {
+  Entries = 'Entries',
+  Memories = 'Memories',
```
karesti, line 216: This is not "Memories" but the Memory entries.

karesti (changes requested):

src/app/Caches/DataDistribution.tsx
```diff
@@ -0,0 +1,148 @@
...
+    const [isOpenStatsOptions, setIsOpenStatsOptions] = useState<boolean>(false);
+    const [statsOption, setStatsOption] = useState<string>(DataDistributionStatsOption.Entries);
+
+    const { dataDistribution, loading, error } = useDataDistribution(props.cacheName);
+
+    const onSelectStatsOptions = (event, selection, isPlaceholder) => {
+        if(selection === t('caches.cache-metrics.data-distribution-option-entries'))
+            setStatsOption(DataDistributionStatsOption.Entries);
+        else if(selection === t('caches.cache-metrics.data-distribution-option-memory'))
+            setStatsOption(DataDistributionStatsOption.Memories);
```
karesti, line 35: we are passing the hardcoded enum here instead of the label. this should be done the other way around

karesti (changes requested):

jabolina:
> @dpanshug I added more comments. we are almost there. Please check with @jabolina to add the memory property. Memory is not the total entries in memory, but the memory that the node is consuming from a JVM perspective. Does this make sense @jabolina ? This is not blocking for the PR. We can add the memory feature (like the threshold one) later

IIUC, the memory in this PR refers to the number of entries in memory, and the total includes memory + disk. The memory from the JVM perspective is in the cluster endpoint, which returns the total memory available per node.

karesti:
Great @jabolina ! is there any possibility to know the memory per cache too?

jabolina:
@karesti The cache stats endpoint exposes an estimation in bytes, with the JSON in [here](https://infinispan.org/docs/dev/titles/rest/rest.html#rest_v2_cache_detail) is the value `$.stats.data_memory_used`. I think we could extend that to the distribution endpoint too.

src/app/Caches/DataDistribution.tsx
dpanshug: For  enum: 
```
export enum DataDistributionStatsOption {
  Entries = 'Entries',
  Memories = 'Memorie entries',
}
```

For translation:
```      
"data-distribution-option-entries": "Total number of entries",
"data-distribution-option-memory": "Total memory used",
```
The values in both are different, can't we use "Total number of entries" in the enum instead of "Entries"?

dpanshug:
> > @dpanshug I added more comments. we are almost there. Please check with @jabolina to add the memory property. Memory is not the total entries in memory, but the memory that the node is consuming from a JVM perspective. Does this make sense @jabolina ? This is not blocking for the PR. We can add the memory feature (like the threshold one) later
> 
> IIUC, the memory in this PR refers to the number of entries in memory, and the total includes memory + disk. The memory from the JVM perspective is in the cluster endpoint, which returns the total memory available per node.

@jabolina can you specify which cluster endpoint has memory?

jabolina:
@dpanshug This one https://infinispan.org/docs/dev/titles/rest/rest.html#rest_v2_cluster_distribution It should work similarly to the cache distribution endpoint.

dpanshug:
> @dpanshug This one https://infinispan.org/docs/dev/titles/rest/rest.html#rest_v2_cluster_distribution It should work similarly to the cache distribution endpoint.

This returns `memory_available` and `memory_used`, are we looking for total memory? available+used?

jabolina:
@dpanshug  Yes, total memory = available + used.

karesti:
@dpanshug @jabolina I will create an issue to add the memory part in the REST API and then use it in the console as well.
This JIRA will only contain entries in memory and total number of entries

@dpanshug needs rebase from main

karesti:
@jabolina @dpanshug the console issue, and the rest api issue in blocked by https://issues.redhat.com/browse/ISPN-14130

src/app/Caches/DataDistribution.tsx
karesti: I will check the PR and I tell you

karesti:
Screenshot 2022-09-14 at 15 12 01
Screenshot 2022-09-14 at 15 12 12

karesti (changes requested):
Labels fix and then ready to merge

src/app/Caches/DataDistribution.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    containerComponent={<ChartVoronoiContainer 
+                                            labels={({ datum }) => datum.y !== 0 ? `${datum.y}` : `${t('caches.cache-metrics.data-distribution-no-entry')}`} 
+                                            constrainToVisibleArea />}
+                    domain={{ y: [0, maxDomain] }}
+                    domainPadding={{ x: [30, 25] }}
+                    legendData={[{ 
+                        name: statsOption === DataDistributionStatsOption.Entries ? 
+                                        t('caches.cache-metrics.data-distribution-option-entries')
+                                        : t('caches.cache-metrics.data-distribution-option-memory')}, { 
+                        name: 'Total memory used above/below threshold'
```
karesti, line 87: Please remove this label as I explained in my comment.

karesti (changes requested):

src/services/infinispanRefData.ts
```diff
@@ -212,6 +212,6 @@ export enum ConfigDownloadType {
 }
 
 export enum DataDistributionStatsOption {
-  Entries = 'Entries',
-  Memories = 'Memories',
+  TotalEntries = 'Total number of entries',
```
karesti, line 215: OK

src/services/infinispanRefData.ts
```diff
@@ -212,6 +212,6 @@ export enum ConfigDownloadType {
...
+  MemoryEntries = 'Total number of entries in memory',
```
karesti, line 216: "Entries in memory"

karesti (changes requested):

karesti (approved):

PR #232, opened 2022-07-19
ISPN-13894 Alert when protobuf schemas are updated

dipguptaredhat:
https://issues.redhat.com/browse/ISPN-13894

Warning Alert is used, https://www.patternfly.org/v4/components/alert

### DIFF
--- a/src/app/ProtoSchema/ProtobufSchemasDisplay.tsx
+++ b/src/app/ProtoSchema/ProtobufSchemasDisplay.tsx
@@ -1,4 +1,4 @@
-import React, {useEffect, useState} from 'react';
+import React, { useEffect, useState } from 'react';
 import {
   Alert,
   AlertVariant,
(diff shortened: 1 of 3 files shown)


dipguptaredhat:
Find one bug.

During editing a protobug, when we clear the content inside Schema and save it, it displays a spinner and further we can't edit to correct that. 

Need to have a validation before saving the edited schema? @karesti 

Screenshot from 2022-07-19 09-07-27

src/app/ProtoSchema/ProtobufSchemasDisplay.tsx
```diff
@@ -69,24 +69,24 @@ const ProtobufSchemasDisplay = (props: {
...
-                initSlice,
-                initSlice + schemasPagination.perPage
-              )
-            );
-          } else {
-            setError(eitherResponse.value.message);
-          }
-        }).then(() => setLoading(false));
-      }
+    if (loading) {
```
karesti, line 72: could we take the opportunity to refactor this part and use a hook ?

karesti (changes requested):

dipguptaredhat:
@karesti added a commit, to fetch schemas from hook.
The issue I am getting is when we add new schema the list is not getting updated.

src/app/ProtoSchema/ProtobufSchemasDisplay.tsx
```diff
@@ -69,25 +69,18 @@ const ProtobufSchemasDisplay = (props: {
...
+      props.setProtoSchemasCount(schemas.length);
+      const initSlice =
+        (schemasPagination.page - 1) * schemasPagination.perPage;
+      setFilteredSchemas(
+        schemas.slice(
+          initSlice,
+          initSlice + schemasPagination.perPage
+        )
+      )
+    }
```
dipguptaredhat, line 82: we need to setLoading(false) here, but how can we set that after if loop.

dipguptaredhat (commented):
@karesti added a commit, to fetch schemas from hook.
The issue I am getting is when we add new schema the list is not getting updated.

karesti (changes requested):
Fix the error message (those can be seen in the dev tools in the console tab)

TextArea.js?6e2c:36 TextArea: TextArea requires either an id or aria-label to be specified

The hook should replace all the ConsoleServices calls, in every component.

dipguptaredhat:
> The hook should replace all the ConsoleServices calls, in every component.

When replacing this `protobufService.getSchema(schemaName).` with the hook, it's throwing an error `Error: Rendered more hooks than during the previous render.`

Maybe because this hook is called up multiple times, for all the schemas.

dipguptaredhat (commented):
> The hook should replace all the ConsoleServices calls, in every component.

When replacing [this](https://github.com/infinispan/infinispan-console/blob/b192fe58edd3c2b5c21bf4fe43b16b1873ff2756/src/app/ProtoSchema/ProtobufSchemasDisplay.tsx#L92) `protobufService.getSchema(schemaName).` with the hook, it's throwing an error `Error: Rendered more hooks than during the previous render.`

Maybe because this hook is called up multiple times, for all the schemas.

karesti (changes requested):
leave the hook use for all the calls out of this PR
however, the proto-schema.cy.js  is failing

andyuk1986 (approved):
Tests are passing now and I have checked all works fine. Thanks Dipanshu :)

PR #249, opened 2022-09-09
ISPN-14092 Cache Configuration Wizard Direct Link

dpanshug:
https://issues.redhat.com/browse/ISPN-14092


dpanshug:
On choosing "Cache wizard" in the sidebar "Data container" also shows active

src/app/routes.tsx
```diff
@@ -117,10 +117,10 @@ const routes: IAppRoute[] = [
   {
     component: CreateCache,
     exact: true,
-    label: 'Caches',
+    label: 'Cache Wizard',
     path: '/container/:cmName/caches/create',
```
dpanshug, line 121: "cmName" not fetching from Router.

src/app/routes.tsx
```diff
@@ -117,10 +117,10 @@ const routes: IAppRoute[] = [
   {
     component: CreateCache,
     exact: true,
-    label: 'Caches',
+    label: 'Cache Wizard',
```
karesti, line 120: @dvagnero I don't like the title but I'm unable to think in a better name. Any ideas ?

src/app/CacheManagers/CacheTableDisplay.tsx
```diff
@@ -583,9 +583,13 @@ const CacheTableDisplay = (props: {
   };
 
   const buildCreateCacheButton = () => {
-    if (!ConsoleServices.security().hasConsoleACL(ConsoleACL.CREATE, connectedUser)) {
-      return '';
-    }
+
+    // Will remove comment when the feature is ready
```
karesti, line 587: What is this comment ?

karesti (changes requested):

src/app/CacheManagers/CacheTableDisplay.tsx
dpanshug: Will remove that :+1:

src/app/routes.tsx
dvagnero: Hey @karesti , I think the term wizard is widely known but here it's wizard + the possibility to provide a cache configuration. So I'm thinking about "Cache Configuration" or  "Cache Setup". Later today I'm giving demo to Erin from the UX team so I'll check with her

dpanshug: Hey @dvagnero any update on the name?

dvagnero: Apologies for the delay. I did mention this change but didn't follow up on it. @edonehoo do you have any suggestions?

edonehoo: I was actually going to suggest "Cache Setup" before seeing that you offered that idea! I think it's best to give context to what the wizard is used for, which "setup" seems to cover. "Create Cache" may also work, if clicking on that tab will always initiate the creation of a new cache (rather than continuing a cache setup that was previously started).

dpanshug: I also like "Cache setup", if everyone agrees moving ahead with that :+1:

src/app/routes.tsx
karesti: I added a fix. We don't need the cache manager name in the url in that case, but other cases use it too. We need to do a proper solution to grab the cache manager information and pass it around. probably with state providers and hooks. 

For now I think we can use what I've done in the commit. Please check it and rebase/ merge / format the stuff properly. Probably we need a "Loading"  generic component too that could be re used around as the error one.

karesti (changes requested):
Properly put things together with the commit I did

PR #257, opened 2022-09-22
ISPN-14155 Cross on rebalance popup window doesn't work

dpanshug:
https://issues.redhat.com/browse/ISPN-14155


andyuk1986 (approved):
Tested locally. Works fine now.

PR #267, opened 2022-10-14
ISPN-14231 Upgrade to PTF4 2022.13

karesti:
https://issues.redhat.com/browse/ISPN-14231


dpanshug:
Do we also need to review component upgrades? https://www.patternfly.org/v4/developer-resources/release-notes

karesti:
> Do we also need to review component upgrades? https://www.patternfly.org/v4/developer-resources/release-notes/#components

no, we check tests and a global review for the console and that's all

PR #284, opened 2022-11-16
ISPN-14309 Improve Counters Table Filtering

dpanshug:
https://issues.redhat.com/browse/ISPN-14309
https://marvelapp.com/prototype/8d66729/screen/69367847


karesti (changes requested):
Screenshot 2022-11-16 at 23 20 12

The select box should not be displaying any number, and add a filter icon like in the cache filter. Filters should not be mandatory. For consistency, I think we should be doing the same behaviour as with caches: displaying a filter.

Screenshot 2022-11-16 at 23 20 17

I think we could improve the message, and we can do the same on caches in another issue, displaying a message where we say there are no counters with the chosen filters. And if there are no counters at all, then display the same message as on the caches but adapted to counters "Select "Create a cache" and provide configuration in XML, JSON, or YAML format or use a custom template. You can also create caches from the CLI and remote clients."

dpanshug:
By default are we displaying all counters (strong + weak) together in the table?  Because in the design create button is specific to one type of counter.

image

src/app/CacheManagers/CounterTableDisplay.tsx
```diff
@@ -128,21 +107,116 @@ const CounterTableDisplay = (props: { setCountersCount: (number) => void; isVisi
...
+    } else {
+      rows = counters;
+      setActions(
+        selectedFilters.counterType === CounterType.STRONG_COUNTER ? strongCountersActions : weakCountersActions
+      );
+    }
+    setRows(rows);
+  };
+
+  const onSelectFilter = (event, selection) => {
```
dpanshug, line 125: @karesti I've used a lot of if-else for this function, let me if there is any refracting possible for this.

src/app/assets/languages/en.json
```diff
@@ -118,6 +104,24 @@
...
+      "initial-value": "Initial value",
+      "lower-bound": "Lower bound:",
+      "upper-bound": "Upper bound:",
+      "concurrency-level": "Concurrency level:",
+      "storage": "Storage",
+      "counter-configuration": "Configuration",
+      "counter-type": "Counter type",
+      "no-counters-status": "No counters yet",
+      "no-counters-body": "Select \"Create counter\" or you can also create counters from the CLI and remote client.",
+      "no-filtered-counter-body": "No counters match the filter."
```
dpanshug, line 124: @dvagnero we need a body saying "The filters you have chosen doesn't have any counters" or like that. Please check if this work or we need to change.

dvagnero: @dpanshug I'd say "No counters match your filter criteria. Click \"Create counter\" and configure a counter."
Do you plan to add this message to filters for Caches, Tasks etc?

dvagnero:
Hi @dpanshug I'm sorry for the delay. I have few docs suggestions can I add my commit?

dpanshug:
> Hi @dpanshug I'm sorry for the delay. I have few docs suggestions can I add my commit?

Yes @dvagnero you can push your commit, thanks

src/app/assets/languages/en.json
```diff
@@ -118,6 +104,24 @@
...
+      "current-value": "Current value",
+      "initial-value": "Initial value",
+      "lower-bound": "Lower bound:",
+      "upper-bound": "Upper bound:",
+      "concurrency-level": "Concurrency level:",
+      "storage": "Storage",
+      "counter-configuration": "Configuration",
+      "counter-type": "Counter type",
+      "no-counters-status": "No counters yet",
+      "no-counters-body": "Select \"Create counter\" or you can also create counters from the CLI and remote client.",
```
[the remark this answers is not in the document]
dpanshug: i would prefer "Create a counter"

dpanshug:
image
@dvagnero in table, counter storage is in all caps (VOLATILE/PERSISTENT), should we present that in Capitalization (Volatile/Persistent) ?

src/services/infinispanRefData.ts
```diff
@@ -226,6 +226,6 @@ export enum CounterType {
 }
 
 export enum CounterStorage {
-  VOLATILE = 'VOLATILE',
```
dpanshug: @dvagnero We need to change filtering function if we change from all-caps to Capitalization in CounterStorage.

Do you recommend not using all-caps?

src/services/infinispanRefData.ts
```diff
@@ -226,6 +226,6 @@ export enum CounterType {
...
-  PERSISTENT = 'PERSISTENT'
+  VOLATILE = 'Volatile',
```
dpanshug, line 229: @dvagnero We need to change filtering function if we change from all-caps to Capitalization in CounterStorage.

Do you recommend not using all-caps?

src/app/assets/languages/en.json
dpanshug: Yes we may add this similar for other things also

src/services/infinispanRefData.ts
dvagnero: Sorry, I didn't know. On my end it changed capitalization for the counter filter and not inside the table. 
image
Is there any reason for capitalization in this case? What are you trying to achieve? 
In general, capitalize the abbreviations and optionally some error/warning (like WARNING: Do not ....) . Capitalizing every letter in other cases is perceived as screaming at the user :) 
If you want to highlight the info you might want to change the style, make it bold

dpanshug: Actually, the variable from backend we are receiving is in all-caps.
So that's why I was previously using VOLATILE/PERSISTENT.

Now if you check with the recent changes filtering with counter storage is not working. We need to change filtering function.

dpanshug: Anyways, I also prefer your way.

Making necessary changes :+1:

dvagnero: You're right filtering based on counter storage doesn't work

dpanshug: > You're right filtering based on counter storage doesn't work

I've added new commit, now it's working

dpanshug: @dvagnero are you making any other changes?

dvagnero: I'm done with the changes. From docs perspective it looks good

dpanshug:
@karesti you can have a final review and we merge

karesti (approved):

PR #292, opened 2022-12-01
[14.x] More tests are implemented 

andyuk1986:
covering cache details page (entries add/update/delete/searching by key and value), cache creation wizard without/with template including all available options in wizard,

data container page with filtering/hiding/showing/deleting caches.

2 tests with cache creation without template are commented at the moment as there is an issue with inserting a text into textarea appeared after cypress upgrade to the latest version. I am working on that.

After merge I will cherry pick the changes and will make a PR to main branch.


cypress/e2e/cache-create-wizard-template.cy.js
```diff
@@ -19,11 +19,71 @@ describe('Cache Creation Wizard', () => {
...
+    cy.contains('Cache ' + cacheName + ' successfully created with e2e-test-template.');
     // Once the cache created, redirection to main page is done and the cache should be visible
     //Is redirected to Data Container page
     cy.get('#cluster-manager-header').should('exist');
     cy.get('[data-cy=cacheManagerStatus]').should('exist');
     cy.get('[data-cy=rebalancingSwitch]').should('exist');
     cy.contains(cacheName);
   })
+
+//   it.only('successfully creates without a template a JSON config', () => {
```
karesti, line 31: @andyuk1986 are these tests going to be back non commented?

andyuk1986: @karesti yes, at the moment I have an issue with cypress typing JSON/XML to cache creation textarea. These tests were passing with Cypress 10.4.x but then after upgrade to 10.8.x and now to 11.x.x they are failing because while typing the cypress "eats" symbols. As soon as I find the solution (or maybe it is a bug on cypress side, not sure), I will uncomment the tests.

karesti (approved):

PR #305, opened 2023-01-03
Counters 

dpanshug:
Delete a counter: https://issues.redhat.com/browse/ISPN-14298
Reset counter: https://issues.redhat.com/browse/ISPN-14303
Strong counter creation: https://issues.redhat.com/browse/ISPN-12080
Weak counter creation: https://issues.redhat.com/browse/ISPN-12081
Delta validation: https://issues.redhat.com/browse/ISPN-14456

### DIFF
--- a/src/app/assets/languages/en.json
+++ b/src/app/assets/languages/en.json
@@ -132,18 +132,18 @@
       "reset-action": "Reset",
       "modal-reset-title": "Reset counter",
       "create-action": "Create",
-      "modal-create-title": "Create counter",
+      "modal-create-title": "Create a counter",
       "modal-lower-bound": "Lower bound",
       "modal-upper-bound": "Upper bound",
       "modal-concurrency-level": "Concurrency level",
       "modal-counter-name": "Counter name",
-      "modal-storage": "Storage",
+      "modal-storage": "Counter storage",
       "modal-strong-counter": "Strong counter",
       "modal-weak-counter": "Weak counter",
-      "modal-initial-value-invalid": "Initial value must be between lowerbound and upperbound (inclusive)",
+      "modal-initial-value-invalid": "Initial value must be between lower bound and upper bound (inclusive)",
       "modal-lower-bound-invalid": "Lower bound must be lesser than upper bound",
-      "modal-storage-tooltip": "Storage tooltip",
-      "modal-counter-type-tooltip": "Counter type tooltip"
+      "modal-storage-tooltip": "{{brandname}} can save counter values in a persistent storage so the data is available when clusters restart. If you select a volatile storage, counter values are permanently deleted when clusters restart or stop",
+      "modal-counter-type-tooltip": "The value of a strong counter is stored in a single key for consistency. {{brandname}} recommends using strong counters when the counter's value is needed after each update or when a bounded counter is needed. The value of a weak counter is stored in multiple keys. The weak counter is suitable for use cases where the result of the update operation is not needed or the counter’s value is not required too often  "
     }
   },
   "caches": {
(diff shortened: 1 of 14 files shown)


karesti (changes requested):
Several issues 

**Create Counter:**

* When we have radio buttons, one value needs to be selected by default
* Add popover tooltip like in the wizard form to explain What means storage Persistent/Volatile
*  Add popover tooltip to explain diff between strong and weak
* Initial value should be a number field, I can add anything
* We need a validation before trying to create the counter and do the alert

"Unexpected error creating counter dd "ISPN029508: Invalid initial value for counter. It must be between 22 and 22 (inclusive) but was 11" 
Before leaving the create form, we need those validations.

** Add Delta **
* I can add any delta even when I have upper bound and lower bound.
* The popup should contain the name of the counter since we don't know which counter we are changing once we open it, like remove.

** Reset **
The wording on the reset should contain more information about the counter. We don't know which is the  counter name or the values that will be reset, even if we do have that information.

There is something weird when we manipulate (add/remove). The page does not seem to get refreshed well. 
I need to dig more in the code details, I did more a behaviour review at this point.

src/app/utils/counterUtils.ts
```diff
@@ -0,0 +1,26 @@
+import { CounterType } from '@services/infinispanRefData';
+
+export function createCounterConfig(counterConfig) {
```
karesti, line 3: this can be unit tested

dvagnero:
Hi @dpanshug , I pushed one commit with docs suggestions. I'll take a look at the messages about invalid values tomorrow.
Two things I noticed.

- During counter creation I don't get a green check for the upper value 
image

- When I try to create a counter that has the same name as already existing counter I get a confirmation message but no counter is created.

karesti:
test failing now @dpanshug

dvagnero:
Added another commit @dpanshug

src/app/assets/languages/en.json
```diff
@@ -140,8 +140,8 @@
       "modal-storage": "Counter storage",
       "modal-strong-counter": "Strong counter",
       "modal-weak-counter": "Weak counter",
-      "modal-initial-value-invalid": "Initial value must be between lower bound and upper bound (inclusive)",
-      "modal-lower-bound-invalid": "Lower bound must be lesser than upper bound",
+      "modal-initial-value-invalid": "The initial value must be between the lower bound and the upper bound inclusively",
```
dpanshug, line 143: Only the upper bound is inclusive.

dpanshug (commented):
@dvagnero Thanks for reviewing.

src/app/assets/languages/en.json
```diff
@@ -125,9 +125,25 @@
...
+      "create-action": "Create",
+      "modal-create-title": "Create a counter",
+      "modal-lower-bound": "Lower bound",
+      "modal-upper-bound": "Upper bound",
+      "modal-concurrency-level": "Concurrency level",
+      "modal-counter-name": "Counter name",
+      "modal-storage": "Counter storage",
+      "modal-strong-counter": "Strong counter",
+      "modal-weak-counter": "Weak counter",
+      "modal-initial-value-invalid": "The initial value must be between the lower bound and the upper bound inclusively",
```
dpanshug, line 143: Only the upper bound is inclusive.

dpanshug:
> Hi @dpanshug , I pushed one commit with docs suggestions. I'll take a look at the messages about invalid values tomorrow. Two things I noticed.
> 
> * During counter creation I don't get a green check for the upper value
>   image
> * When I try to create a counter that has the same name as already existing counter I get a confirmation message but no counter is created.

For the validation of upper bound, we already have validation for lower bound which states that it should be lesser than upper bound and similar for initial value also. 

image

And for the same counter name issue, I have created a JIRA for improving the API [ISPN-14468](https://issues.redhat.com/browse/ISPN-14468)

karesti (approved):

PR #322, opened 2023-02-22
ISPN-12106 Add a refresh button in the cache detail

dpanshug:
https://issues.redhat.com/browse/ISPN-12106


karesti:
Screenshot 2023-03-06 at 16 57 29

karesti (changes requested):
The icon has no label, it's not accessible that way and is hard to understand what it does.
We need a real button eventually with the icon if you want

https://www.patternfly.org/v4/components/button#small-buttons

dpanshug:
Please review changes so I can refresh backport PR, thanks @karesti

karesti (approved):

PR #324, opened 2023-02-23
Cypress and Unit Tests are added for refreshed Counters page.

andyuk1986:


src/app/Counters/DeleteCounter.tsx
```diff
@@ -41,7 +42,7 @@ const DeleteCounter = (props: {
         >
           {t('cache-managers.counters.delete-action')}
         </Button>,
-        <Button key={'Cancel'} aria-label={'Cancel'} variant={ButtonVariant.link} onClick={props.closeModal}>
+        <Button key={'Cancel'} aria-label={'Cancel'} variant={ButtonVariant.link} onClick={props.closeModal} data-cy="cancelButton">
```
karesti, line 45: I'm wondering if can't use the Key or aria label or other instead of data-cy for this tests?
Maybe also naming more explicitely "cancelButtonCreateCounter" or something like that will be better. Since cancelButton is very generic

karesti (changes requested):
small comment on naming of data-cy and related to the use of this as well. but probably is ok using it as it is and free of patternfly stuff that might change

src/app/Counters/ResetCounter.tsx
```diff
@@ -39,7 +40,7 @@ const ResetCounter = (props: {
         >
           {t('cache-managers.counters.reset-action')}
         </Button>,
-        <Button key={'Cancel'} aria-label={'Cancel'} variant={ButtonVariant.link} onClick={props.closeModal}>
+        <Button key={'Cancel'} aria-label={'Cancel'} variant={ButtonVariant.link} onClick={props.closeModal} data-cy="cancelButton">
```
karesti, line 43: @andyuk1986 same here for the renaming please

karesti (changes requested):
a few renamings better IMO

src/app/Counters/ResetCounter.tsx
andyuk1986: Done :)

karesti (approved):

PR #334, opened 2023-03-31
ISPN-14230 Download server report from cluster membership list

dpanshug:
* Action added to download server report.
* Table changed to TableComposable
* Labels exported

https://issues.redhat.com/browse/ISPN-14230

### DIFF
--- a/src/app/ClusterStatus/ClusterStatus.tsx
+++ b/src/app/ClusterStatus/ClusterStatus.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { useEffect, useState } from 'react';
 import {
+  Button,
   Bullseye,
   Card,
   CardBody,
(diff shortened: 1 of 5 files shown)


src/app/ClusterStatus/ClusterStatus.tsx
```diff
@@ -21,108 +21,65 @@ import {
...
+    name: t('cluster-membership.node-name'),
+    physicalAdd: t('cluster-membership.physical-address')
+  };
+
+  const clusterMemberActions = (row): IAction[] => [
     {
-      title: 'Physical address'
+      title: t('cluster-membership.download-report'),
+      onClick: () => {
+        ConsoleServices.server().downloadReport(row.name);
```
karesti, line 52: can;t this be done with a hook ?
would be nice to try to unit test with jest mocking the server report download calls

src/services/serverService.ts
```diff
@@ -21,4 +21,15 @@ export class ServerService {
   public async getVersion(): Promise<Either<ActionResponse, string>> {
     return this.utils.get(this.endpoint, (data) => data.version);
   }
+
+  /**
+   * Get server report for a given nodeName
+   */
+  public downloadReport(nodeName: string): void {
```
karesti, line 28: server errors are not handled there

karesti: is this the way it works in Patternfly ?

src/app/ClusterStatus/ClusterStatus.tsx
dpanshug: Added hook and unit test.

src/services/serverService.ts
dpanshug: Changed function to handle server error. 
There's no mention of the download feature in patternfly.

dpanshug:
One thing I've noticed, downloaded reports from server `./run-server-for-e2e.sh` are not extractable 
image

Whereas it's working fine with server `15.0.0-SNAPSHOT` 

@andyuk1986 can you check pls.

karesti:
@jabolina any clue with @dpanshug issue ?

andyuk1986:
Let me check it out

andyuk1986:
@dpanshug @karesti seems that when we run the maven command to install the latest console changes to the server, i.e. build the console and copy the dist to the static directory of the already running server, then the generated report is corrupted. But if we will restart the server after that action, then the server is generated normally. Maybe we need to make some changes to script and maven commands so that later when tests are running we get normally generated report?

dpanshug:
> @dpanshug @karesti seems that when we run the maven command to install the latest console changes to the server, i.e. build the console and copy the dist to the static directory of the already running server, then the generated report is corrupted. But if we will restart the server after that action, then the server is generated normally. Maybe we need to make some changes to script and maven commands so that later when tests are running we get normally generated report?

Yes, after restarting server reports are extractable.

But the other thing, after comparing reports from both server `15.0.0-SNAPSHOT` and `./run-server-e2e`, server `./run-server-e2e` lacks some files and folder, as you can see in screenshot below

(left: `./run-server-e2e`, right: `15.0.0-SNAPSHOT`)
Screenshot from 2023-04-12 12-20-28

Is this okay? @jabolina

karesti:
The feature is ok from the console, but we need a release of Infinispan Dev in 15 for testing

karesti:
integration testing needs to be done with a released server

jabolina:
The missing files/folders can happen when some of the commands are not available in the system, or, like in this case, if it is unable to copy some of the folders from the server's root. That can happen because of file ownership, for example. +1 for Katia's suggestion of running on a released server.

karesti (commented):
LGTM but we need a server release so will be on hold now

karesti:
@dpanshug @andyuk1986 the server Dev01 has been released so we can change the version in main, 14.0 branch will remain with 14.0 server

karesti (changes requested):
use 15.0.0.Dev01 server here

dpanshug:
Feature is not working with the released 15.0.0.Dev01 :roll_eyes: 
@jabolina can you confirm if the API is released?

jabolina:
I was able to use it on 15.0.0.Dev01 image.

```bash
$ docker pull infinispan/server:15.0.0.Dev01
$ docker run -it -p 11222:11222 -e USER="admin" -e PASS="password" infinispan/server:15.0.0.Dev01
```

On the other terminal:

```bash
$ curl --digest -u admin:password -O 'http://127.0.0.1:11222/rest/v2/server/report/f439e0909dbd-59934'
```

The `f439e0909dbd-59934` was the name my server received. The name is printed on the docker output.

karesti (changes requested):
works well
the only issue I see is that there is no notification when the download has started.
I click the button and there is some time before the download is happening, so I don't know if it worked or not.
I would add an alert to say the download is happening (green) or that has failed for any reason with the error altert.

src/app/ClusterStatus/ClusterStatus.tsx
```diff
@@ -195,18 +143,63 @@ const ClusterStatus: React.FunctionComponent<any> = (props) => {
...
+              ) : (
+                filteredClusterMembers.map((row) => {
+                  return (
+                    <Tr key={row.name}>
+                      <Td dataLabel={columnNames.name}>{row.name}</Td>
+                      <Td dataLabel={columnNames.physicalAdd}>{row.physical_address}</Td>
+                      <Td dataLabel={columnNames.physicalAdd}>
+                        <Button
+                          variant="link"
+                          isInline
```
karesti, line 190: add 
```xml
 icon={!downloading ? <DownloadIcon/> : null}
```

karesti (changes requested):

karesti (approved):

PR #340, opened 2023-04-14
fix translation not found warning messages

karesti:


dpanshug:
@karesti We have used translation in `aria-label` in many different places also, do we need to remove them too?

karesti:
> @karesti We have used translation in `aria-label` in many different places also, do we need to remove them too?
if you run the console, you will see that any translation message that does not exist, get's a warning log. those need to be removed.
the aria label is an accessibility property
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label
we need it in many places, translated, but not everywhere and not as a selector element probably. we need to check the overrall use of it, specially in those places where the translation is applied now. 
the ones I removed there don't exist and log warnings

dpanshug:
Got it, I'll make sure to factor this in next time.

karesti:
@dpanshug you can merge this

PR #341, opened 2023-04-14
XSite Repl is configured and new tests are added.

andyuk1986:


karesti:
we can close these

karesti:
won't do

PR #342, opened 2023-04-18
ISPN-13909 Special symbols handling during Protobuf schema creation

dpanshug:
https://issues.redhat.com/browse/ISPN-13909


dpanshug:
@andyuk1986 `rbac_func` Cypress test is failing, because the test is checking for `contains19 Caches` but only has 16 in `data/caches`

karesti (approved):

PR #338, opened 2023-04-11
ISPN-14761 Displays encoding message for Java and JBoss marshalling

karesti:
https://issues.redhat.com/browse/ISPN-14761

PR #344, opened 2023-04-19
ISPN-14682 Do not retrieve entries when no encoding in cache

dpanshug:
https://issues.redhat.com/browse/ISPN-14682

### DIFF
--- a/src/app/providers/CacheDetailProvider.tsx
+++ b/src/app/providers/CacheDetailProvider.tsx
@@ -3,6 +3,7 @@ import { ConsoleServices } from '@services/ConsoleServices';
 import { useConnectedUser } from '@app/services/userManagementHook';
 import { ConsoleACL } from '@services/securityService';
 import { ContentType } from '@services/infinispanRefData';
+import { isEncodingAvailable } from '@app/utils/encodingUtils';
 
 const initialContext = {
   error: '',
@@ -66,7 +67,7 @@ const CacheDetailProvider = ({ children }) => {
               })
               .finally(() => {
                 setLoading(false);
-                setLoadingEntries(true);
+                isEncodingAvailable(cache) && setLoadingEntries(true);
               });
           } else {
             setError(maybeCm.value.message);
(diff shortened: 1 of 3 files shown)


src/app/providers/CacheDetailProvider.tsx
```diff
@@ -66,7 +66,7 @@ const CacheDetailProvider = ({ children }) => {
               })
               .finally(() => {
                 setLoading(false);
-                setLoadingEntries(true);
+                cache.queryable && setLoadingEntries(true);
```
karesti, line 69: this is not ok. the cache being queriable is not the solution

karesti (changes requested):

karesti:
@dpanshug check here https://github.com/infinispan/infinispan-console/pull/338

No encoding means no entries tab at all.

There are cases that the entries fetch can fail. try creating caches with all the encodings and putting data on them

src/app/providers/CacheDetailProvider.tsx
```diff
@@ -40,6 +40,13 @@ const CacheDetailProvider = ({ children }) => {
     }
   };
 
+  const isEncodingAvailable = (): boolean => {
```
karesti, line 43: export this to the file 

the Empty encoding is an encoding type enum. use ||, not and
can be inlined in a single like

```
return cache.encoding.key  ... ||  ... 
```

karesti (changes requested):

src/app/utils/encodingUtils.ts
```diff
@@ -0,0 +1,3 @@
+export function isEncodingAvailable(cache: DetailedInfinispanCache): boolean {
+  return cache.encoding.key !== 'Empty' || cache.encoding.value !== 'Empty';
```
dpanshug, line 2: @karesti let me know, if this okay, so I can update backport

karesti: you can use the enum

karesti (approved):

karesti:
error seems unrelated so I will merge but we need to fix this :)

PR #346, opened 2023-04-27
Encoding utils error

dpanshug:
Resolving TypeError from encodingUtils file 
image (1)

PR #347, opened 2023-04-27
[14.0] Encoding utils error

dpanshug:
backporting #346


andyuk1986 (approved):
The related tests after the fix are passing on my local.

PR #355, opened 2023-05-10
[14.0] ISPN-14856 Do not show Isolation locking property when not transactional cache

dpanshug:
https://issues.redhat.com/browse/ISPN-14856
Back-porting #354


andyuk1986:
Works as expected. Thanks and merged!

PR #359, opened 2023-05-19
[14.x]  JDG-6220 CVE-2023-0842 xml2js: node-xml2js: xml2js is vulnerable to prototype pollution [jdg-8]

dpanshug:
https://issues.redhat.com/browse/JDG-6220

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,17909 @@
 {
   "name": "infinispan-console",
   "version": "14.0.7",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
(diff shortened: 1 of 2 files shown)


karesti (approved):

pruivo:
@karesti @dpanshug we need to ship licenses to be legally compliant. If you remove `license-reporter`, please add another tool to generate the licenses.xml and licenses.html. Thanks!
This PR will be reverted in the product.

karesti:
@pruivo damm, we need this fix for product too. Dipanshu asked around and we thought this was ok. 
is the licenses mandatory for each component?

karesti:
@pruivo we might be good, looks like we are not using the package dependency to generate licenses but the maven front plugin instead?

PR #366, opened 2023-06-05
Fixing tests after ISPN-14718 is implemented.

andyuk1986:
I have fixed the console tests which were failing after ISPN-14718 was implemented.


andyuk1986:
Would be great if this PR is merged today, so that is already available in tomorrow's CD binaries.

dpanshug (approved):
LGTM :)

PR #373, opened 2023-06-28
ISPN-14966 Connected clients page

dpanshug:
https://issues.redhat.com/browse/ISPN-14966

### DIFF
--- a/src/app/utils/formatAge.ts
+++ b/src/app/utils/formatAge.ts
@@ -1,0 +1,36 @@
+/**
+ * Formats a given date string into a friendly age format.
+ * The friendly age format includes hours, minutes, and seconds.
+ * @param dateString The date string to format.
+ * @returns The formatted age string.
+ */
+export function formatAge(dateString: string): string {
+  const date = new Date(dateString);
+  const now = new Date();
+
+  if (date > now) {
+    return ''; // Return an empty string for future dates
+  }
+
+  const diff = Math.abs(now.getTime() - date.getTime());
+
+  const hours = Math.floor(diff / (1000 * 60 * 60));
+  const minutes = Math.floor((diff / (1000 * 60)) % 60);
+  const seconds = Math.floor((diff / 1000) % 60);
+
+  let formattedAge = '';
+
+  if (hours > 0) {
+    formattedAge += hours + ' hr';
+  }
+
+  if (minutes > 0) {
+    formattedAge += ' ' + minutes + ' min';
+  }
+
+  if (seconds > 0) {
+    formattedAge += ' ' + seconds + ' s';
+  }
+
+  return formattedAge.trim();
+}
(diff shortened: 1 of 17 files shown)


src/app/utils/formatAge.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Formats a given date string into a friendly age format.
+ * The friendly age format includes hours, minutes, and seconds.
+ * @param dateString The date string to format.
+ * @returns The formatted age string.
+ */
+export function formatAge(dateString: string): string {
```
karesti, line 7: there is a library already included that does this formatting I think, no need to do it by coding a function again, but I'm not sure.

karesti: if you add functions like this, you need a unit test to cover as well

src/services/serverService.ts
```diff
@@ -28,4 +28,11 @@ export class ServerService {
   public async getLatestVersion(): Promise<Either<ActionResponse, string>> {
```
karesti, line 28: Unrelated, but I spotted this

The version we are using in the Server should not be grabbed from the github API
There is an endpoint (about uses it) that displays the version of the server we are using.
Is misleading, IMO, putting this version like current version as the latest one, and won't work for DG
IMO this call should be removed and use the server version API

karesti (commented):
Screenshot 2023-07-03 at 18 36 08

The endpoint is unavailable still as I see (running dev01)
 I will review the behaviour from main when these comments will be addressed

karesti:
Screenshot 2023-07-04 at 09 34 51

karesti:
- Display principal N/A, never display "null"
- connection names are not accurate, this should be fixed not tu duplicate names
- if client version is null, display N/A 
- Principal tab should be bigger

@tristantarrant this is what I'm getting running the console
-

src/services/serverService.ts
```diff
@@ -28,4 +28,11 @@ export class ServerService {
...
     return this.utils.get('https://api.github.com/repos/infinispan/infinispan/releases/latest', (data) => data.name);
   }
+
+  /**
+   * Get connected clients
+   */
+  public async getConnectedClients(): Promise<Either<ActionResponse, ConnectedClients[]>> {
+    return this.utils.get(this.endpoint + '/connections', (data) => data);
```
karesti, line 36: filter duplicates here please

karesti (changes requested):

src/services/serverService.ts
dpanshug: Removed in #377

src/app/utils/formatAge.ts
dpanshug: Can't find a library that directly does this formatting. Also added a unit test for this.

PR #375, opened 2023-06-30
ISPN-14962 Metrics tab

dpanshug:
https://issues.redhat.com/browse/ISPN-14962

Previously, we did not display the memory used in the chart, now it's available with cache with eviction by memory.
Also changed options to select type of data distribution.


src/services/cacheConfigUtils.ts
```diff
@@ -463,4 +463,24 @@ export class CacheConfigUtils {
...
+  /**
+   * Function to check if cache has eviction by memory
+   * Recursively checks if memory exists as key in object
+   * @param config
+   */
+  public static checkMemoryExists(config: unknown): boolean {
+    if (typeof config === 'object' && config !== null) {
+      if ('memory' in config) {
+        return true;
+      } else {
```
karesti, line 476: else can be removed and chain, since there is the return

karesti (changes requested):
small code change, but more about look and feel

karesti:
Screenshot 2023-07-03 at 18 14 08
The labels look very big for me. would not be better to use an smaller font ? 
Also I believe with approx. abbreviate labels are shorter too so we wont have the sized to be that weird

karesti:
Screenshot 2023-07-03 at 18 18 28
numbers are likely to be bigger that the labels on the right that have fixed size. Put more size for the number part

dpanshug:
The numbers are emphasized with a bolder font compared to the labels.
image

Also, the labels are designed to be concise, with the shortest possible size of <p>. If we decrease the size further, we would need to use <small> which is lighter and less easily visible.
image

karesti:
@dpanshug there is a formatter for numbers

dpanshug:
@karesti how this looks?

Screenshot from 2023-07-10 12-33-00

karesti:
@dpanshug looks well but still a couple of things:

1) the approximate is too big. Could be use the abbreviation for this? I think Approx. is enough.
2) the Data access part, I would like to see a screenshot with very big numbers, because this will be the case. Those numbers must be formatted using the format utility.

dpanshug:
@karesti 

Numbers are formatted using `displayUtils.formatNumber()`

image

karesti (approved):

PR #376, opened 2023-07-03
ISPN-14920 Counters tab display

dpanshug:
https://issues.redhat.com/browse/ISPN-14920


karesti (changes requested):
The counters tab does not get refresh when we create a new counter

karesti:
Screenshot 2023-07-04 at 10 07 10

dpanshug:
Creating the first counter and then fetching, the issue seems to be from the backend, discussed this with Pedro R. and created jira [ISPN-15023](https://issues.redhat.com/browse/ISPN-15023).

karesti:
@dpanshug when I move to another page, the fetch works correctly when I come back to the cache manager page

dpanshug:
> @dpanshug when I move to another page, the fetch works correctly when I come back to the cache manager page

Yes, the counter is fetched after reloading the page. When we create the very first counter on the server, it is not fetched. However, if we create and delete the first counter and then create another counter on the same server, the fetch operation works correctly.

The issue is with the very first counter on the server.

karesti:
> Creating the first counter and then fetching, the issue seems to be from the backend, discussed this with Pedro R. and created jira [ISPN-15023](https://issues.redhat.com/browse/ISPN-15023).

I fixed temporarly by calling twice when the result is empty

src/app/CacheManagers/CacheTableDisplay.tsx
```diff
@@ -506,6 +506,30 @@ const CacheTableDisplay = (props: { cmName: string; setCachesCount: (count: numb
     </MenuToggle>
   );
 
+  const handleClickOutside = (event: MouseEvent) => {
```
karesti, line 509: @dpanshug this allows the filter to be closed when you click outside

cypress/e2e/counters.cy.js
```diff
@@ -150,93 +150,95 @@ describe('Counters CRUD', () => {
         cy.get("[aria-label='counters-table-label']").contains("td", "strong-5").should("not.exist");
       });
 
-      it ('successfully filters counters', () => {
```
karesti, line 153: @andyuk1986 the behaviour works, but this test stops because is unable to open "the filter". i'm under the impression there is a bug in patternly when doing the responsive design behavior ... do you have time to check for this? (you can say not now and I'll log a jira)

karesti:
failures unrelated

PR #377, opened 2023-07-04
Removing latest version feature

dpanshug:


karesti (changes requested):
Display the server version, the running one, not the one in github

karesti (approved):

PR #455, opened 2024-05-20
French translation

ryanemerson:
@karesti As part of the day of learning I decided to play with some hugging face models translation models. The following is the result of an automated script that converts the `en.json` from this repository into french via the inference API of [this](https://huggingface.co/Helsinki-NLP/opus-mt-en-fr) Apache 2.0 license model.

As you know, my French is very limited, so I'll let you be the judge of the translation quality. If it seems reasonable, then we can easily modify my script to convert to other languages :slightly_smiling_face:


src/app/assets/languages/fr.json
```diff
@@ -0,0 +1,1011 @@
...
+      "modal-grant-title": "Accorder un nouvel acc\u00e8s",
+      "modal-grant-description": "Accorder l'acc\u00e8s \u00e0 {{brandname}} \u00e0 un mandant en accordant des r\u00f4les.",
+      "modal-principal-name": "Directeur",
+      "modal-principal-name-is-required": "Le nom du mappeur de r\u00f4le principal est requis",
+      "modal-roles-is-required": "Choisir au moins un r\u00f4le",
+      "modal-principal-exists": "Le r\u00f4le principal de la carte {{name}} existe d\u00e9j\u00e0",
+      "modal-user-exists": "Un utilisateur nomm\u00e9 {{name}} existe. Vous n'avez pas besoin de cr\u00e9er un mapper principal.",
+      "modal-roles": "R\u00f4les",
+      "modal-roles-list-placeholder": "S\u00e9lectionner les r\u00f4les",
+      "grant-success": "Acc\u00e8s accord\u00e9 \u00e0 {{112,114,105,110,99,105,112,97,108,78,97,109,101}.",
```
ryanemerson, line 988: The CSV of numbers inbetween `{{ }}` is a bug in my script. I'll fix this if you think the previous translations are reasonable, otherwise I'll just close the PR.

tristantarrant:
Maybe @yrodiere can have a look

src/app/assets/languages/fr.json
```diff
@@ -0,0 +1,1011 @@
...
+    "develop": "Applications de code pour diff\u00e9rents cas d'utilisation.",
+    "tutorials": "Ex\u00e9cuter des tutoriels rapides",
+    "go-to-console": "Ouvrir la console"
+  },
+  "login-form": {
+    "login-main": "Connectez-vous \u00e0 {{brandname}}.",
+    "login-label": "{{brandname}} connexion.",
+    "login": "Connectez-vous",
+    "cancel": "Annuler",
+    "dialog-label": "{{brandname}} Ouvrir une session dans la bo\u00eete de dialogue.",
```
yrodiere, line 58: Hard to say without context, but if this is the title of a login modal, this translation is very incorrect.

That being said, even in the english version, mentioning "dialog" seem weird... ?

```suggestion
    "dialog-label": "Connexion à {{brandname}}",
```

src/app/assets/languages/fr.json
```diff
@@ -0,0 +1,1011 @@
...
+    "description": "Cette page n'existe plus.",
+    "button": "Retour \u00e0 la page d'accueil"
+  },
+  "not-auth-page": {
+    "title": "Acc\u00e8s non autoris\u00e9",
+    "description": "L'utilisateur connect\u00e9 n'a pas la permission ADMIN.",
+    "button": "Retour \u00e0 la page d'accueil"
+  },
+  "support": {
+    "no-user": "Pouvoirs requis.",
```
yrodiere, line 75: Very weird wording.

But then the original "Credentials required" is weird as well, isn't it? It probably should have been "Authentication required"?

```suggestion
    "no-user": "Authentification requise.",
```

src/app/assets/languages/fr.json
```diff
@@ -0,0 +1,1011 @@
...
+    "no-user-text": "Vous devez saisir des identifiants pour acc\u00e9der \u00e0 la console. Cr\u00e9ez un nom d'utilisateur et un mot de passe pour continuer.",
+    "no-user-label": "Aucune identification utilisateur n'existe.",
+    "reload-button": "Recharger",
+    "text-create-user": "Cr\u00e9er un nom d'utilisateur et un mot de passe, puis recharger.",
+    "text-command": "bin/cli.sh user create myuser -p changeme -g admin"
+  },
+  "cache-managers": {
+    "title": "R\u00e9cipient de donn\u00e9es",
```
yrodiere, line 83: "Récipient" is more what you would call a bowl of a vase.

I think people would expect a more straightforward translation here: `container` => `conteneur`.

```suggestion
    "title": "Conteneur de donn\u00e9es",
```

src/app/assets/languages/fr.json
```diff
@@ -0,0 +1,1011 @@
...
+    "caches-tab": "Caches",
+    "counters-tab": "Compteurs",
+    "tasks-tab": "T\u00e2ches",
+    "schemas-tab": "Sch\u00e9mas",
+    "cache-table-label": "Caches",
+    "cache-name": "D\u00e9nomination",
```
yrodiere, line 89: "Dénomination" is very formal, something you'd find on official documents rather than in a web UI.

The simpler translation of "Name" is just "Nom".

```suggestion
    "cache-name": "Nom",
```

src/app/assets/languages/fr.json
```diff
@@ -0,0 +1,1011 @@
...
+    "cache-mode": "Type",
+    "cache-health": "Sant\u00e9",
+    "cache-features": "Caract\u00e9ristiques",
```
yrodiere, line 92: Technically this means "Characteristics". Not sure that's what is meant by "Features" here: "Fonctionnalités" might be a better fit. Again, hard to say without context.

src/app/assets/languages/fr.json
```diff
@@ -0,0 +1,1011 @@
...
+    "cache-status": "Le pr\u00e9sent r\u00e8glement entre en vigueur le jour suivant celui de sa publication au Journal officiel de l'Union europ\u00e9enne.",
```
yrodiere, line 93: Serious hallucination here: an empty string got translated to "This regulation is effective the day after its publication to the official journal of the European union".

src/app/assets/languages/fr.json
```diff
@@ -0,0 +1,1011 @@
...
+    "ignore": "Masquer",
+    "undo-ignore": "Afficher",
+    "ignored-status": "Cach\u00e9",
+    "delete": "Supprimer",
+    "available": "Mise \u00e0 disposition",
```
yrodiere, line 98:
```suggestion
    "available": "Rendre disponible",
```

yrodiere (commented):
I had a look for lines 1 to 100, but I can't spend the time needed to process the almost 1000 lines, sorry :)

Note to @ryanemerson: this is a very nice thing you've done here, so I want to clarify I'm not commenting on your work, but on the AI's translation.

I'd say you're getting more or less what you'd expect from an AI: good in general, unnatural from time to time, completely nonsensical in some cases (see the empty string translated to something about the European Union, near the end).

As a primarily French speaker who also happens to speak English, I think I'd personally prefer the original version in English, rather than end up with awkward or sometimes misleading wording.

I suppose this could be useful to people who have a very bad understanding of English, but then... I suspect these people will have a hard time finding good documentation, so they are unlikely to be able to use Infinispan in the first place.

You seem to already have other translations, though, so perhaps you've already decided that an imperfect translation is better than no translation; I'll admit there are good points on both sides of the argument.

ryanemerson:
> I had a look for lines 1 to 100, but I can't spend the time needed to process the almost 1000 lines, sorry :)

Thanks so much for the detailed feedback you have provided, it's really illustrated the limitations of a purely AI based solution :slightly_smiling_face: 

> As a primarily French speaker who also happens to speak English, I think I'd personally prefer the original version in English, rather than end up with awkward or sometimes misleading wording.
> 
> I suppose this could be useful to people who have a very bad understanding of English, but then... I suspect these people will have a hard time finding good documentation, so they are unlikely to be able to use Infinispan in the first place.

This is a very good point!

> You seem to already have other translations, though, so perhaps you've already decided that an imperfect translation is better than no translation; I'll admit there are good points on both sides of the argument.

I was hoping that we could generate something that was imperfect, but usable, however given that we can't guard against hallucinations without a human-review this was probably too optimistic. I think for this to be effective we would need to combine the generated AI with a thorough human review, which given that this isn't a priority and we only have a small number of languages spoken in the team, is not realistic.

PR #486, opened 2024-09-11
ISPN-16412 Display cache config when not Healthy

karesti:
https://issues.redhat.com/browse/ISPN-16412


karesti:
Depends on https://github.com/infinispan/infinispan/pull/12851

PR #570, opened 2025-01-31
[15.0] Fixing keycloak config for running with KC 25+

andyuk1986:
Backport for #569


karesti:
@andyuk1986 looks like there is an issue with the server

